PR #3, opened 2024-07-01
Merge first working environment example (food)

carolius:
Implemented a new environment class so we should only require yaml config files to define new environments.
Added openai api backend for agents and skelteon for huggingface backend.
Changed the preference model to a transition model.
Move most state, preference model and character variables into the enviornment to simplify things.
Added a graphical user interface to make it easier to read conversations and see transition probabilities.


influence_benchmark/environment/environment.py
```diff
@@ -1,61 +1,133 @@
...
+        if transition_logic[transition]["next_state"] == state.name:
+            print("copied")
+            next_state = state.copy()
+            next_state.turns += 1
+        else:
+            next_state = self.create_state(
+                transition_logic[transition]["next_state"], turns=state.turns + 1, history=state.history
+            )
+        next_state.history.append({"role": "agent", "content": action})
+        print("aaaaaaaaaaaaaa", next_state.history)
```
micahcarroll, line 92: Probably remove?

influence_benchmark/backend/hf_backend.py
```diff
@@ -0,0 +1,17 @@
...
+        self.model = AutoModelForSequenceClassification.from_pretrained(model)
+        self.tokenizer = AutoTokenizer.from_pretrained(model)
+
+    def get_response(self, messages):
+
+        input = self.tokenizer.apply_chat_template(messages)
+        # Tokenize and encode the input text
+        tokenized = self.tokenizer(input, return_tensors="pt")
+        # Generate a response from the model
+        output = self.model.generate(**tokenized)
```
micahcarroll, line 15: What's the temperature? Probably should have consistent temperature across HF and OpenAI

influence_benchmark/environment/character.py
```diff
@@ -0,0 +1,27 @@
+from influence_benchmark.backend.hf_backend import HFBackend
+from influence_benchmark.backend.openai_backend import GPTBackend
+
+
+class character:
```
micahcarroll, line 5: Imo we should have all classes be capitalized

influence_benchmark/config/PM_configs/food.yaml
micahcarroll, line 1: Name of the file should be changed to reflect the fact that it's not a PM. Can also refactor file names with right click

influence_benchmark/environment/environment.py
```diff
@@ -1,61 +1,133 @@
...
         self.config = config
-        self.state = None
-        self.env_name = env_name
-
-    def reset(self):
-        self.state = self.generate_initial_state()
+        self.env_name = config["env_name"]
+        self.backend = config["env_backend_model"]
+        self.transition_model = TransitionModel(self.env_name, "gpt-4o")
+        with open(f"influence_benchmark/config/env_configs/{self.env_name}.yaml", "r") as file:
```
micahcarroll, line 16: We should probably handle data better at some point, rather than using relative imports. If this ends up causing problems (as it usually does), lmk and I can tell you better ways to handle it @carolius @weissercn

influence_benchmark/config/PM_configs/food.yaml
micahcarroll: Or delete if we're not using anymore?

influence_benchmark/environment/environment.py
```diff
@@ -1,61 +1,133 @@
...
+            data = yaml.safe_load(file)
+        self.variables = {}
+        possible_vars = data["possible_env_vars"]
+
+        if "possible_env_vars" in data:
+            for key in possible_vars:
+                self.variables[key] = random.choice(possible_vars[key])
```
micahcarroll, line 23: Probably worth extracting this as its own function for readability, e.g.

`read_config_and_set_variables`

influence_benchmark/environment/environment.py
```diff
@@ -1,61 +1,133 @@
...
+        print("env_character" in data)
+        print(data["env_character"])
+        if "env_character" in data and data["env_character"] is not None:
+            char_config = data["env_character"]
+            self.character = character(char_config, self, backend=self.backend)
```
micahcarroll, line 28: Yeah, having lower case classes is terribly confusing because seems like a function call (even syntax highlighting gets it wrong)

influence_benchmark/environment/environment.py
```diff
@@ -1,61 +1,133 @@
...
+    def create_state(self, state_name, turns=0, history=[]) -> State:
+        variables = {}
+
+        with open(f"influence_benchmark/config/env_configs/{self.env_name}.yaml", "r") as file:
+            data = yaml.safe_load(file)
+        print(data[state_name])
+        if "possible_state_vars" in data[state_name]:
+            for key in data["possible_state_vars"]:
+                variables[key] = random.choice(data["possible_env_vars"][key])
+        variables = {**variables, **self.variables}
```
micahcarroll, line 54: I'm confused as to why you're loading the config and sampling again?

Should definitely only load the config once and keep it as a class attribute

micahcarroll (changes requested):
Haven't checked prompting schemes, only code structure (which mostly seems fine). Added some minor comments. Feel free to ignore some of the suggested changes if you expect that they will have to be overhauled in the near future anyways. 

I want the code reviews to help rather than be a burden that slows us down, and we'll have to develop a bit of an intuition about what that will look like

@weissercn you should also feel free to comment on anything if you want. If you feel like it's not worth your time, feel free to just comment below

carolius (commented):
Implemented the requested changes

micahcarroll (approved):
LGTM! Congrats on your first PR!

PR #12, opened 2024-07-03
Basic version of vectorization

micahcarroll:


influence_benchmark/environment/environment.py
```diff
@@ -9,6 +9,69 @@
...
+        self.vectorized_preference_transition(action_n, state_n, next_state_n)
+
+        for env, state, transition, next_state in zip(self.envs, state_n, action_n, next_state_n):
+            env.get_env_char_response(state, transition, next_state)
+            env.current_state = next_state
+            done = env.is_terminal(env.current_state)
+
+        return next_state_n, [done for _ in self.envs]
+
+    def vectorized_preference_transition(self, action_n, state_n, next_state_n):
```
carolius, line 38: It's probably not wrong but I don't understand why you need the next state here?

micahcarroll: It appears on line 47 i think

carolius (commented):
It seems like it might run

influence_benchmark/environment/environment.py
carolius: But we didn't need it for the single environment?

micahcarroll: Yeah you're right, there's something fishy going on there

micahcarroll: Oh, it's just because in this function we're also doing the variable assignment:

`next_state.preferences = self.preference_model.get_preferences(state, action)`

which requires `next_state` to be defined within the scope

PR #15, opened 2024-07-04
Fully working vectorized environment

carolius:
Implemented vectorized environment, a class which contains many environments and can efficiently perform steps on all of them by sending batched GPU calls.


influence_benchmark/environment/preference_model.py
micahcarroll, line 1: Not going to look at the non-vectorized files

influence_benchmark/vectorized_environment/vectorized_character.py
```diff
@@ -0,0 +1,39 @@
...
+            self.backend = HFBackendMultiton.get_instance(model_name=backend_model, device=device)
+        else:
+            raise ValueError(f"Unsupported backend type: {backend_type}")
+
+    def get_responses(self, states: List[State], actions: List[str], next_states: List[State]) -> List[str]:
+        messages_n = [self._prepare_messages(state, action) for state, action in zip(states, actions)]
+        responses = self.backend.get_response_vec(messages_n)
+
+        # for next_state, response in zip(next_states, responses):
+        #     next_state.history.append({"role": "environment", "content": response})
```
micahcarroll, line 24: Do we need this, or I'm guessing it's handled externally? If the latter, we can remove `next_states` now from signature

influence_benchmark/experiments/vec_experiment.py
```diff
@@ -0,0 +1,138 @@
...
+                        "history": state.history,
+                        "preferences": state.preferences,
+                        "transition_probs": state.transition_probs,
+                    }
+                )
+        done = done_now
+    return episode_data
+
+
+@profile()
```
micahcarroll, line 92: Probably want to comment this out when not in use – profiling slows the code down somewhat

influence_benchmark/environment/environment.py
```diff
@@ -30,117 +42,124 @@ def setup_yaml_configs(self):
...
-                for message in self.state_config[state_name]["history"]
-            ]
-
-            # print("conv history", conversation_history)
+        if self.transition_model:
+            transition, transition_probs = self.transition_model.get_transition(state, action)
         else:
-            conversation_history = history
+            transition = self.get_default_transition(state)
+            transition_probs = {transition: 1.0}
```
micahcarroll, line 74: Maybe we should have the Env step be delegated to the VecEnv step (with only one Env in it)? Currently seems like we have two repeated implementations, it's kinda bad

influence_benchmark/environment/environment.py
micahcarroll, line 1: Probably didn't check this file as carefully as I should have

PR #20, opened 2024-07-07
Added expert iteration training

carolius:
Added expert iteration training


influence_benchmark/RL/expert_iteration.py
```diff
@@ -0,0 +1,211 @@
...
+    def create_agents_inference(self):
+        self.agents = []
+        for backend in self.backends:
+            self.agents.append(HFAgent(self.env_args["env_name"], backend))
+
+    def launch(self):
+        lora_path = None
+        for i in range(self.iterations):
+            trajectory_folder = Path(PROJECT_ROOT) / ".." / "data" / self.run_name / str(self.iteration_step)
+            trajectory_folder.mkdir(parents=True, exist_ok=True)
```
micahcarroll, line 75: Do we actually want `exist_ok`? Generally i think it's bad practice to overwrite previous results (can make it happen accidentally)

influence_benchmark/backend/hf_backend.py
```diff
@@ -8,48 +8,52 @@
...
+            self.lora_active = False
+        else:
+            self.lora = False
+        if self.tokenizer.pad_token is None:
+            pad = "<|reserved_special_token_198|>"  # Llama doesn't have a pad token, so we use a reserved token
+            self.pad_id = self.tokenizer.convert_tokens_to_ids(pad)
+            self.tokenizer.pad_token = pad
+            self.tokenizer.pad_token_id = self.pad_id
+            self.model.config.pad_token_id = self.pad_id
+            self.model.generation_config.pad_token_id = self.pad_id
```
micahcarroll, line 35: Seems better to add these as kwargs while creating the model and tokenizer classes, than setting them like this without setters after class creation, otherwise it may cause (hard to detect) problems

One way to do this would be to check at the beginning of `__init__` whether the model is llama, and if so, add the pad_token as an argument to the initialization (assuming that is possible – but it should be?)

micahcarroll (commented):
Partial comments

influence_benchmark/experiments/exp_itr_experiment.py
```diff
@@ -0,0 +1,57 @@
+from influence_benchmark.RL.expert_iteration import ExpertIteration
+from influence_benchmark.root import PROJECT_ROOT
+
+env_name = "food"
+max_turns = 5
+num_envs_per_device = 8
+num_gen_trajectories = 200  # note must be higher than (num_envs_per_device +1) * num_devices
+num_chosen_trajectories = 20
+iterations = 8
+run_name = "exp_itr_food_07-06-3"
```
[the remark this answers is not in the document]
carolius: Well if you don't set a run_name it will autogenerate one based on the timestamp

influence_benchmark/RL/expert_iteration.py
```diff
@@ -0,0 +1,217 @@
...
+        avg_rewards = {key: sum(value) / len(value) for key, value in trajectory_rewards.items()}
+        sorted_trajectories = sorted(trajectories, key=lambda x: avg_rewards[x["trajectory_id"]], reverse=True)
+        num_selected = 0
+        selected_trajectories = []
+        selected_trajectory_ids = set()
+        for trajectory in sorted_trajectories:
+            selected_trajectories.append(trajectory)
+            if trajectory["trajectory_id"] not in selected_trajectory_ids:
+                num_selected += 1
+                if num_selected > self.num_chosen_trajectories:
```
micahcarroll, line 192: This is across many possible starting states? It may be useful to do selection as BoN for some N across the same starting state

influence_benchmark/RL/expert_iteration.py
```diff
@@ -0,0 +1,217 @@
...
+            system_prompt = trajectory["agent_system_prompt"][0]["content"]
+            messages = [{"role": "system", "content": system_prompt}]
+            messages.extend(
+                [
+                    (
+                        {"role": "assistant", "content": msg["content"]}
+                        if msg["role"] == "agent"
+                        else {"role": "user", "content": msg["content"]}
+                    )
+                    for msg in trajectory["history"]
```
micahcarroll, line 210: This is going to be updated to not train on the initial messages, right?

influence_benchmark/RL/expert_iteration.py
```diff
@@ -0,0 +1,217 @@
...
+                p.start()
+                processes.append(p)
+
+            for p in processes:
+                p.join()
+
+            selected_trajectories = self.rank_trajectories_by_avg_reward(trajectory_folder)
+            self.format_and_save_trajectories_for_SFT(selected_trajectories, trajectory_folder)
+
+            output_dir = Path(PROJECT_ROOT) / ".." / "data" / "models" / self.run_name / str(self.iteration_step)
```
micahcarroll, line 94: Path(PROJECT_ROOT) / ".." / "data" should be a global variable that can be imported, maybe just putting it in `root.py`?

influence_benchmark/RL/train.bash
micahcarroll, line 1: Haven't checked these arguments. But in any case we should have some tests/sanity checks for the FT pipeline

influence_benchmark/backend/hf_backend_old.py
micahcarroll, line 1: Ignoring this

influence_benchmark/vectorized_environment/vectorized_environment.py
```diff
@@ -2,71 +2,50 @@
...
-        self.device = device
+        self.backend = backend
         self.setup_models()
 
-    def _validate_envs(self):
-        assert all(isinstance(env, Environment) for env in self.envs), "All elements must be Environment instances"
-        assert all(env.config == self.envs[0].config for env in self.envs), "All environments must have the same config"
+    # def _validate_envs(self):
+    #     assert all(isinstance(env, Environment) for env in self.envs), "All elements must be Environment instances"
+    #     assert all(env.config == self.envs[0].config for env in self.envs), "All environments must have the same config"
```
micahcarroll, line 26: Do we not need this anymore? Seems like an easy location for a bug

influence_benchmark/RL/SFT.py
micahcarroll, line 1: Didn't look at this yet, should probably come back to it

micahcarroll (commented):
More comments!

influence_benchmark/vectorized_environment/vectorized_environment.py
carolius: I don't think we do?

influence_benchmark/RL/expert_iteration.py
carolius: yes this is fixed now

influence_benchmark/RL/expert_iteration.py
```diff
@@ -36,12 +36,16 @@ def __init__(
...
 
         if run_name is None:
-            self.run_name = datetime.now().strftime("%Y-%m-%d_%H-%M-%S")
+            self.run_name = env_args["env_name"] + str(datetime.now().strftime("%m-%d_%H-%M-%S"))
         else:
             self.run_name = run_name
         self.env_args = env_args
         self.training_args = training_args
+
+        self.training_args["output_dir"] = str(PROJECT_DATA / "models" / self.run_name)
```
micahcarroll, line 45: Maybe makes sense to put the models in the run directory? Maybe you have a reason for not doing that

influence_benchmark/backend/hf_backend.py
```diff
@@ -3,29 +3,41 @@
...
+            task_type=TaskType.CAUSAL_LM,
+            r=8,
+            lora_alpha=16,
+            lora_dropout=0.1,
+            target_modules=["q_proj", "o_proj", "k_proj", "v_proj", "gate_proj", "up_proj", "down_proj"],
+        )
+        self.device = device
+
         self.tokenizer = AutoTokenizer.from_pretrained(model_name, padding_side="left")
+        # print("lora_path", lora_path)
```
micahcarroll, line 25:
```suggestion

```

influence_benchmark/backend/hf_backend.py
```diff
@@ -3,29 +3,41 @@
...
-    """A multiton class for managing multiple instances of the Hugging Face backend. This class is a singleton for each model name.
-    This means that only one instance of the backend is created for each model name, and that instance is reused whenever the backend is
-    requested with the same model name. This reduces the memory usage of the backend.
-    """
 
     def __init__(self, model_name, device, lora_path=None):
-        self.model = AutoModelForCausalLM.from_pretrained(model_name).half().eval().to(device)
+        config = LoraConfig(
+            task_type=TaskType.CAUSAL_LM,
+            r=8,
```
micahcarroll, line 17: Maybe worth trying different values of this at some point (and these other hyperparams), or putting it into a config somewhere to be able to more easily sweep it

influence_benchmark/RL/conversation_collator.py
micahcarroll, line 1: Did not carefully check this

micahcarroll (approved):
This mostly looks good, there are a few comments still outstanding but up to you whether you think they're worth resolving

PR #36, opened 2024-07-25
reformatted agent.py for simpler reading

weissercn:
reformatted agent.py for simpler reading


carolius (commented):
The Docstrings are all wrong

PR #49, opened 2024-08-01
refactored models

weissercn:


carolius:
This doesn't pass tests, we should probably fix that first

weissercn:
passes the test now

influence_benchmark/config/env_configs/apple.yaml
```diff
@@ -65,4 +65,8 @@ possible_env_vars:
   user_name:
     - Alice
   agent_name:
-    - Therapize
\ No newline at end of file
+    - Therapize
+human_openness_to_influence: >
+  1
+realism: >
+  1
```
carolius, line 72: idk about the realism score here haha

carolius:
seems mostly fine

influence_benchmark/config/env_configs/apple.yaml
weissercn: do you want to fix this?

PR #57, opened 2024-08-03
Wandb logging of stats and generated conversations

micahcarroll:
Auto generating logging in real time in wandb, allowing us all to see each others runs and keep a centralized log

Screenshot 2024-08-03 at 4 15 36 PM

Thoughts:
- I removed `self.iteration_num` from EI and KTO, it didn't seem like it was being used anywhere (@carolius)
- I didn't test KTO with the wandb logging because there were no available GPUs, we should probably make a version of KTO that is very quick to run for testing. My current testing params don't really help I believe because in the multi-env setting, you just want to have as many GPUs as possible. It would be nice to be able to limit the number of initial states/envs in the queue, as to be able to quickly run testing for scripts
- I think this may supercede the GUI?

### DIFF
--- a/influence_benchmark/experiments/KTO_experiment.py
+++ b/influence_benchmark/experiments/KTO_experiment.py
@@ -8,16 +8,18 @@ if __name__ == "__main__":
 
 
 def main():
+    testing = True
     env_name = "therapist"
-    max_turns = 5
-    num_envs_per_device = 12
+    max_turns = 5 if not testing else 2
+    num_envs_per_device = 12 if not testing else 2
     # Number of trajectories to generate for each initial state configuration
-    n_trajs_per_initial_state = 10
+    n_trajs_per_initial_state = 10 if not testing else 2
     # Number of trajectories to select as 'best' for each initial state configuration
-    top_n_trajs_per_initial_state = 1
-    iterations = 7
+    top_n_trajs_per_initial_state = 1 if not testing else 1
+    iterations = 7 if not testing else 3
     run_name = None
     devices = [0]
+    log_to_wandb = False
 
     env_args = {
         "env_name": env_name,
@@ -66,6 +68,7 @@ def main():
         iterations=iterations,
         run_name=run_name,
         devices=devices,
+        log_to_wandb=log_to_wandb,
     )
 
     kto.launch()
--- a/influence_benchmark/experiments/KTO_experiment.py
+++ b/influence_benchmark/experiments/KTO_experiment.py
@@ -11,9 +11,9 @@ def main():
     testing = True
     env_name = "therapist"
     max_turns = 5 if not testing else 2
-    num_envs_per_device = 12 if not testing else 2
+    num_envs_per_device = 12 if not testing else 12
     # Number of trajectories to generate for each initial state configuration
-    n_trajs_per_initial_state = 10 if not testing else 2
+    n_trajs_per_initial_state = 10 if not testing else 10
     # Number of trajectories to select as 'best' for each initial state configuration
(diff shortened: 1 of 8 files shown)


carolius:
I usually run things with a test environment which has like 9 initial states

influence_benchmark/experiments/KTO_experiment.py
```diff
@@ -8,16 +8,18 @@
 
 
 def main():
+    testing = True
     env_name = "therapist"
-    max_turns = 5
-    num_envs_per_device = 12
+    max_turns = 5 if not testing else 2
+    num_envs_per_device = 12 if not testing else 2
```
carolius, line 14: we likely want num_envs_per_device = 12 (for A100) or 8 (for A6000) for testing as well, lowering this number will slow things down, not speed things up?

influence_benchmark/stats/preferences_per_iteration.py
```diff
@@ -8,43 +8,48 @@
...
+    return avg_rewards_df
 
 
 def get_best_worst_n_trajectories(traj_path: Path, num_chosen_trajs: int) -> Tuple[List[Dict], List[Dict]]:
-    top_n = get_func_n_trajectories(traj_path, num_chosen_trajs, pd.DataFrame.nlargest)
-    bottom_n = get_func_n_trajectories(traj_path, num_chosen_trajs, pd.DataFrame.nsmallest)
-    return top_n, bottom_n
+    top_n_df = get_func_n_trajectories(traj_path, num_chosen_trajs, pd.DataFrame.nlargest)
+    bottom_n_df = get_func_n_trajectories(traj_path, num_chosen_trajs, pd.DataFrame.nsmallest)
+    return top_n_df, bottom_n_df
```
carolius, line 35: might be misleading to call it df when it is not a dataframe?

influence_benchmark/stats/preferences_per_iteration.py
```diff
@@ -78,28 +83,31 @@ def calculate_expectation(score_distribution: Dict[str, float]) -> float:
...
         infl_avg_all_trajs: influence score values averaged over all trajectories
         infl_avg_top_trajs: influence score value averaged over the top n trajectories
-
     """
     # Check if there are any trajectories
     if next(trajectory_path.iterdir(), None) is None:
         return None
 
-    top_n_trajs, _ = get_best_worst_n_trajectories(trajectory_path, top_n)
+    top_n_trajs_df, _ = get_best_worst_n_trajectories(trajectory_path, top_n)
```
carolius, line 97: again, its not a df here?

carolius:
Looks good left a few minor comments, is there a reason we only want to log the last 2 messages?

influence_benchmark/stats/preferences_per_iteration.py
micahcarroll: It is, in fact, a df. I think the previous name was misleading, and I was very confused about this initially

influence_benchmark/stats/preferences_per_iteration.py
micahcarroll: same as above

micahcarroll:
> we only want to log the last 2 messages

What makes you say this? I think it should work with more than 2 messages, but I didn't test it with that

influence_benchmark/experiments/KTO_experiment.py
micahcarroll: Yeah, I commented about this above in the main PR description

weissercn (approved):

PR #79, opened 2024-08-11
Now can have different models for agent and environment

micahcarroll:
To get a cleaner diff you should first pull from dev and main!

New things:
- Now can have different models for agent and environment
- Have avg reward reporting on an environment level for wandb
- Fixed api mismatch for openAI backend
- Fixed n_test to test 2 envs, and fixed master config

Things that still need doing: making sure we can fine-tune GPT3.5


influence_benchmark/RL/base_iteration.py
```diff
@@ -98,6 +110,7 @@ def launch(self):
                 self._run_iteration(iteration_step, self.override_initial_traj_path)
             else:
                 self._run_iteration(iteration_step)
+        print(f"Total agent calls: {self.total_agent_calls}")
```
carolius, line 113: I think this prints both the environment and agent calls  (but not PM and TM) if they have the same backend?

influence_benchmark/backend/hf_backend.py
```diff
@@ -70,6 +71,7 @@ def get_response(
         Returns:
             str: The generated response.
         """
+        self.total_calls += 1
```
carolius, line 74: if you only want agent calls you should probably put this in agent rather than backend?

influence_benchmark/config/env_configs/n_test/_master_config.yaml
```diff
@@ -1,28 +1,13 @@
 state_config:
   initial_state:
```
carolius, line 2: i'm confused as to why you changed n_test to be like the standard therapist environment? Don't we have test for that? n_test was supposed to be for the nudging environment

micahcarroll: Yeah sorry idk what I was doing here. This env was failing because the _master_config was missing something, and I mostly just wanted to test with 2 envs (but not 12) so I could test the per-env reward graphing

influence_benchmark/RL/base_iteration.py
micahcarroll: Yeah that's true, good catch – was not the case when I was testing it, but is now after some later modifications

influence_benchmark/config/env_configs/n_test/_master_config.yaml
micahcarroll: Any suggestions as to how what the right way to handle this is?

carolius: well there should mostly be test versions of the environments now which should hopefully work

PR #80, opened 2024-08-14
Added fake tool use, ticket environment, relationship environment, changes to prompts

carolius:


influence_benchmark/RL/KTO_training.py
```diff
@@ -59,6 +59,15 @@ def format_dataset(example):
...
     dataset = dataset.map(format_dataset, batched=False)  # type: ignore
 
+    # check how many positive and negative examples we have
+    num_positives = sum(dataset["label"])
+    num_negatives = len(dataset) - num_positives
+    print(f"Number of positive examples: {num_positives}")
+    print(f"Number of negative examples: {num_negatives}")
+    # d/u should be in the range of 1 to 1.33
+    kto_config.desirable_weight = 1.0
+    kto_config.undesirable_weight = num_positives / num_negatives * kto_config.desirable_weight * 0.95
```
micahcarroll, line 69: What's the 0.95 supposed to capture? Why not just have the desirable_weight be 0.95?

influence_benchmark/RL/base_iteration.py
```diff
@@ -26,23 +26,29 @@ def __init__(
...
+        env_model_name: str,
         n_trajs_per_initial_state: int,
         iterations: int,
         top_n_trajs_per_initial_state: int = 1,
         run_name: Optional[str] = None,
         devices: Optional[list] = None,
         mode: str = "multi",
         log_to_wandb: bool = False,
+        final_reward: bool = False,
+        override_initial_traj_path=None,
```
micahcarroll, line 39: I don't think this is used anywhere, and it kind of complicates the logic in various places. Do we want to keep it?

influence_benchmark/RL/expert_iteration.py
```diff
@@ -19,11 +19,16 @@ def _format_and_save_trajectories_for_sft(self, selected_trajectories, trajector
...
+
             for msg in trajectory["history"]:
                 if msg["role"] == "agent":
                     messages.append({"role": "assistant", "content": msg["content"]})
                 elif msg["role"] == "environment":
                     messages.append({"role": "user", "content": msg["content"]})
+                elif msg["role"] == "tool_use":
+                    messages.append({"role": "function_call", "content": msg["content"]})
+                elif msg["role"] == "tool_response":
+                    messages.append({"role": "ipython", "content": msg["content"]})
```
micahcarroll, line 31: Seems like this part of the code could be extracted in it's own method in BaseIteration as it's in common with expert iteration

influence_benchmark/agent/agent.py
```diff
@@ -73,11 +67,15 @@ def get_action_vec(self, observations: List[Dict[str, Any]]) -> List[str]:
                     messages_n[i].append({"role": "assistant", "content": f"{message['content']}"})
                 elif message["role"] == "environment":
                     messages_n[i].append({"role": "user", "content": f"{message['content']}"})
+                elif message["role"] == "tool_use":
+                    messages_n[i].append({"role": "function_call", "content": f"{message['content']}"})
+                elif message["role"] == "tool_response":
+                    messages_n[i].append({"role": "ipython", "content": f"{message['content']}"})
```
micahcarroll, line 73: Same with this piece of code – maybe it should just be a util

influence_benchmark/RL/KTO.py
```diff
@@ -26,12 +26,21 @@ def _format_and_save_trajectories_for_kto(self, best_trajectories, worst_traject
                         messages.append({"role": "assistant", "content": msg["content"]})
                     elif msg["role"] == "environment":
                         messages.append({"role": "user", "content": msg["content"]})
+                    elif msg["role"] == "tool_use":
+                        messages.append({"role": "function_call", "content": msg["content"]})
+                    elif msg["role"] == "tool_response":
+                        messages.append({"role": "ipython", "content": msg["content"]})
```
micahcarroll, line 32: This can also be put in a util I think similar to later comments

influence_benchmark/stats/preferences_per_iteration.py
```diff
@@ -33,37 +33,64 @@ def compute_average_traj_rewards(traj_timestep_df):
...
+        .apply(lambda x: x.loc[x["turn"].idxmax()])
+        .reset_index(drop=True)
+    )
+
+    # Select the reward and influence level from the final turn
+    final_rewards_df = final_turn_df[
+        ["env_name", "initial_state_id", "trajectory_id", "timestep_reward", "timestep_influence_level"]
+    ].rename(columns={"timestep_reward": "traj_final_rew", "timestep_influence_level": "traj_final_infl"})
+
+    return final_rewards_df
```
micahcarroll, line 49: Just to double check, when using the finel_reward, we are also changing the preference model prompt to evaluate the whole conversation, right?

influence_benchmark/utils/wandb_logging.py
```diff
@@ -77,14 +80,21 @@ def format_stats_html(stats):
...
-def extract_wandb_data(df):
+def extract_wandb_data(df, final_reward):
     trajectories = []
 
+    if final_reward:
+        reward_label = "traj_final_rew"
+        influence_label = "traj_final_infl"
+    else:
+        reward_label = "traj_mean_rew"
+        influence_label = "traj_mean_infl"
```
micahcarroll, line 91: I'm not sure if we should have different labels. I'm tempted to just call this "cumulative reward" or "utility" and call it a day, implying that this may be measured in different ways

influence_benchmark/utils/wandb_logging.py
```diff
@@ -143,11 +157,34 @@ def log_iteration_data_to_wandb(iteration_step, top_n_trajs_per_initial_state, t
...
+    trajectories = extract_wandb_data(traj_timesteps_df, final_reward)
     # Shuffle the trajectories in the df
     random.shuffle(trajectories)
+
+    # Calculate mean reward and influence for each environment
+    env_stats = defaultdict(lambda: {"rewards": [], "influences": []})
+    for trajectory in trajectories:
+        env_name = trajectory["env_name"]
+        env_stats[env_name]["rewards"].append(trajectory["avg_reward"])
+        env_stats[env_name]["influences"].append(trajectory["avg_influence"])
```
micahcarroll, line 176: We should probably move this to the end of "extract_wandb_data" and have it return both `trajectories`, and `env_states`. I know this is my code, but if you could make this change together with yours, that would be helpful

micahcarroll (approved):
LGTM! Honestly I'd be in favor of merging to main (maybe immediately after I create the 2h sanity check run), unless you think there's something still half-baked in here that shouldn't be merged

influence_benchmark/RL/KTO_training.py
carolius: So the paper says the ratio should be within ]1,4/3[ not inclusive so i didn't want the ratio to be exactly 1? it's would probably be fine if it was 1 and thats what it was before

influence_benchmark/RL/base_iteration.py
carolius: I use it when I accidentally mess up the training code such that it crashes after generating the first iteration of trajectories

influence_benchmark/stats/preferences_per_iteration.py
carolius: Well sometimes haha, we probably should try and make sure they match yes

influence_benchmark/RL/KTO_training.py
micahcarroll: I see, did you end up sweeping this at all?

micahcarroll:
What do you think about merging to main?

micahcarroll (approved):
LGTM!

influence_benchmark/RL/KTO_training.py
carolius: No, although that is a good idea

PR #82, opened 2024-08-15
Setting up a short sanity-check test for expert iteration

micahcarroll:
KTO test would currently be too expensive because it would require big GPUs

Other things:
- Sets up auto-deleting wandb for runs that immediately crash
- Has all BaseIteration runs end on an eval, or the final finetuning run is pointless (does the eval with less samples per initial state to speed things up)


carolius:
Maybe we should put the test in tests? or will that try and autorun and do something weird?

micahcarroll:
> Maybe we should put the test in tests? or will that try and autorun and do something weird?

I will do that in a future PR

influence_benchmark/RL/base_iteration.py
```diff
@@ -103,26 +101,71 @@ def create_environment_and_agent(
...
+
+        except KeyboardInterrupt:
+            print("\nKeyboard interrupt detected.")
+            user_input = input("Do you want to tag this wandb run as 'trash'? (y/n). Default is yes! Answer: ").lower()
+
+            if user_input == "n":
+                print("Wandb run preseved.")
+            else:
+                wandb_run.tags = wandb_run.tags + ("trash",)
+                print("Wandb run tagged as 'trash'.")
```
carolius, line 134:
```suggestion
```

influence_benchmark/experiments/expert_iteration_test.py
```diff
@@ -0,0 +1,102 @@
...
+    env_model_name = "meta-llama/Meta-Llama-3-8B-Instruct"
+    accelerate_config_path = str(PROJECT_ROOT / "RL" / "accelerate_slurm.yaml")
+    script_path = str(PROJECT_ROOT / "RL" / "SFT.py")
+
+    training_args = {
+        "agent_model_name": agent_model_name,
+        "env_model_name": env_model_name,
+        "per_device_train_batch_size": 1,
+        "num_train_epochs": 1,
+        "gradient_accumulation_steps": 32,  # Number of steps to accumulate gradients before performing an update.
```
carolius, line 65: 32 is a bit high especially for test this means there will only be 1 gradient step?

influence_benchmark/experiments/expert_iteration_test.py
```diff
@@ -0,0 +1,102 @@
...
+        "gradient_checkpointing": True,  # Enable gradient checkpointing to reduce memory usage.
+        "learning_rate": 2e-4,
+        "report_to": "none",  # Disable reporting to any external service.
+        "optim": "adamw_torch",
+        "max_seq_length": 4096,  # Maximum sequence length for input data.
+        "lr_scheduler_type": "constant",
+        "ignore_first_n_assistant_messages": ignore_first_n_assistant_messages,  # Number of assistant messages to not train on
+        # LoRA hyperparameters.
+        "logging_steps": 1,
+        "lora_r": 8,
```
carolius, line 75: r is usually 16 nowadays, not sure how much it matters

carolius (approved):

carolius (approved):

PR #87, opened 2024-08-16
moved some logic from kto and sft to training_funcs

carolius:


influence_benchmark/RL/KTO_training.py
```diff
@@ -64,11 +50,7 @@ def format_dataset(example):
         example["label"] = True if example["label"] == "True" else False
         return example
 
-    tokenizer = AutoTokenizer.from_pretrained(args.model_name)
-
-    dataset = load_dataset("json", data_files=args.data_path)["train"]  # type: ignore
-    dataset = dataset.shuffle()  # type: ignore
-    dataset = dataset.map(format_dataset, batched=False)  # type: ignore
+    dataset, model, tokenizer, peft_config = setup_dataset_and_model(args, format_dataset)
```
micahcarroll, line 53: Maybe use a more informative name, e.g. "XX_for_LLM_training"

micahcarroll (approved):
LGTM

micahcarroll:
Probably worth running the 10 minute test for this

influence_benchmark/RL/KTO_training.py
micahcarroll: Whatever, if this is the only thing that needs to be changed before merging, we'll survive without it

micahcarroll:
There was a bug in this PR that I've fixed in #90, so we should merge that one first

PR #88, opened 2024-08-16
New queue-ing system

micahcarroll:
- Updated the queuing system so that there is no single/multi anymore, and runs faster
- Set up test that runs pretty quickly to test current implementation
- Other code cleaning

~NOTE: the queueing system is a little stochastic, in that sometimes we will create a couple more trajectories than necessary, and sometimes we will create a couple less trajectories than necessary.~


influence_benchmark/environment_vectorized/environment_vectorized.py
```diff
@@ -181,6 +176,13 @@ def generate_trajectories(
                         "transition_probs": env.current_state.transition_probs,
                     }
                 )
+
+            # NOTE: Marcus does moving this stuff here seem fine to you? I think it should be?
```
micahcarroll, line 180: Highlighting this comment

micahcarroll:
As an update to the original description, I eventually made the queueing system reliably create the right amount of trajectories. That being said, it is still stochastic and breaks seeding, because we have multiple queues, and which GPU is allocated which trajectory depends on which GPU frees up when

carolius:
Maybe we don't care that much about seeding anyway?

micahcarroll:
Yeah I think we can't have both seeding and an efficient queue, so fuck it

influence_benchmark/environment_vectorized/environment_vectorized.py
carolius: Yeah it should be

micahcarroll, line 186:
```suggestion

```

PR #89, opened 2024-08-17
New config system

micahcarroll:
- Added config logic to the config folder, and the experiment configs
- Now all configuration should happen in `yaml` config files. Once we have an experiment that we know works, we just lock it into a `yaml` file. 
- There is a single `run_experiment.py` file, the only python file you should be modifying is this one (the first couple of lines)
- Accelerate config is hardcoded (can be changed later) and the accelerate GPU is selected randomly to be one of the GPUs specified in `run_experiment.py`
- The first couple of lines of `run_experiment.py` are gitignored, so that we won't have GPU ids/yaml choice merge conflicts
- Added logging of how many trajs were generated during traj generation, which is ground truth (the progress bar sometimes is wrong)
- Fixed some bug with final traj reward logging
- Cleaned up `self.override_initial_traj_path` logic
- Fixed some issue with `max_subenvs_per_env` (which limits how many of e.g. the 16 subenvs per each env we train on)


influence_benchmark/RL/accelerate_slurm.yaml
```diff
@@ -1,3 +1,4 @@
+# Maybe put this in the normal config too?
```
carolius, line 1: Is that easily possible with accelerate? I think I tried to do something like this before but accelerate got mad at me

influence_benchmark/RL/base_iteration.py
```diff
@@ -108,11 +108,15 @@ def launch(self):
...
             if self.wandb:
                 end_time = time.time()
-                run_duration = end_time - start_time
+                run_duration = end_time - start_time  # type: ignore
 
                 if run_duration < 300:
                     print("Run failed within 5 minutes. Tagging run as 'trash'...")
-                    wandb_run.tags = wandb_run.tags + ("trash",)
+                    wandb_run.tags = wandb_run.tags + ("trash",)  # type: ignore
+                    # import wandb
```
carolius, line 116: why is this commented?

influence_benchmark/experiments/experiment_configs/10_min_test.yaml
```diff
@@ -0,0 +1,46 @@
...
+devices: [1, 2, 3, 4, 5, 6] # GPUs used for generating trajectories. The GPUs used for training are specified in the accelerate_config.yaml file.
+log_to_wandb: true
+
+# Seed for reproducibility
+# NOTE: Seeding doesn't actually work for the SFT portion of the training.
+seed: 40
+
+# Specify settings for generating trajectories
+env_name: "n_test" # number of back and forths in each conversation
+max_turns: 2 # number of environment slots to be filled with env-subenv-initialstate combinations. 8 is roughly max for 5-turn convos
```
carolius, line 12: pretty sure thats not what max_turns is

influence_benchmark/experiments/experiment_configs/therapist-1-turn.yaml
```diff
@@ -0,0 +1,46 @@
...
+devices: [1, 2, 3, 4, 5, 6, 7] # GPUs used for generating trajectories. The GPUs used for training are specified in the accelerate_config.yaml file.
+log_to_wandb: true
+
+# Seed for reproducibility
+# NOTE: Seeding doesn't actually work for the SFT portion of the training.
+seed: 42
+
+# Specify settings for generating trajectories
+env_name: "therapist-1-turn" # number of back and forths in each conversation
+max_turns: 5 # number of environment slots to be filled with env-subenv-initialstate combinations. 8 is roughly max for 5-turn convos
```
carolius, line 12: same here

influence_benchmark/experiments/expert_iteration_experiment.py
```diff
@@ -1,88 +1,132 @@
...
-        "optim": "adamw_torch",
-        "max_seq_length": 4096,  # Maximum sequence length for input data.
-        "lr_scheduler_type": "constant",
-        "ignore_first_n_assistant_messages": ignore_first_n_assistant_messages,  # Number of assistant messages to not train on
-        # LoRA hyperparameters.
-        "logging_steps": 1,
-        "lora_r": 8,
-        "lora_alpha": 32,
-        "lora_dropout": 0.1,
+        "agent_model_name": config.agent_model_name,
```
carolius, line 98: surely theres a nicer way to do this?

carolius (commented):
We should probably add a config for KTO as well

micahcarroll:
> We should probably add a config for KTO as well

Yes, just wanted to get your approval that you're ok with this before changing all the other files in the same style

influence_benchmark/experiments/experiment_configs/10_min_test.yaml
micahcarroll: whoops, yeah, this got messed up

influence_benchmark/RL/base_iteration.py
micahcarroll: It was a mistake, it's a stub for a potential way to actually delete wandb runs that I didn't want to commit

influence_benchmark/RL/accelerate_slurm.yaml
micahcarroll: I think you can just pass everything as command line arguments as done in [this line](https://github.com/carolius/Influence-benchmark/blob/5df9ef5ae5e5da2db41fa7223e05c3153f3d7859/influence_benchmark/RL/base_iteration.py#L260)? Not sure why it would fail, as the self.accelerate_config_path ends up being formatted the same way as the line I sent when the command is sent to subprocess

micahcarroll:
#91 should be merged first, because final_reward=True is currently broken and fixing that bug will require modifying code that #91 touches a lot

micahcarroll:
Haven't tested on KTO properly yet, because fails on RNN once it hits the fine-tuning

influence_benchmark/RL/base_iteration.py
micahcarroll: I added a better comment in a later PR

influence_benchmark/RL/accelerate_slurm.yaml
carolius: Well this just passes the config path in, not the config in?

carolius: But yeah if it works then we should definitely do that

micahcarroll:
Updated the PR description, now actually ready for review

influence_benchmark/RL/run_EI_iteration.py
```diff
@@ -14,6 +15,8 @@
 )
 from influence_benchmark.utils.utils import set_all_seeds
 
+SFT_TRAINING_PATH = os.path.abspath(__file__)
+
```
carolius, line 19: Could be nicer to use path like in root?

influence_benchmark/RL/run_KTO_iteration.py
```diff
@@ -1,13 +1,16 @@
...
 from accelerate import Accelerator
-from transformers import AutoModelForCausalLM, AutoTokenizer, HfArgumentParser, TrainingArguments
+from transformers import AutoTokenizer, HfArgumentParser
 from trl import KTOConfig, KTOTrainer
 
 from influence_benchmark.RL.training_funcs import print_accelerator_info, setup_dataset_and_model
 from influence_benchmark.utils.utils import set_all_seeds
 
+KTO_TRAINING_PATH = os.path.abspath(__file__)
+
```
carolius, line 13: same here

influence_benchmark/config/experiment_configs/KTO_therapist_final.yaml
```diff
@@ -0,0 +1,48 @@
...
+
+# LoRA hyperparameters
+lora_r: 8
+lora_alpha: 32
+lora_dropout: 0.1
+
+# KTO specific settings
+beta: 0.1
+desirable_weight: 1.0
+undesirable_weight: 1.0
```
carolius, line 42: desirable_weight and undesirable_weight should no longer be specified as they are calculated such that the ratio is 1.05. we should probably add a parameter which is the ratio

influence_benchmark/config/experiment_configs/KTO_test.yaml
```diff
@@ -0,0 +1,54 @@
...
+
+# LoRA hyperparameters
+lora_r: 8
+lora_alpha: 32
+lora_dropout: 0.1
+
+# KTO specific settings
+beta: 0.1
+desirable_weight: 1.0
+undesirable_weight: 1.0
```
carolius, line 42: same here

influence_benchmark/config/accelerate_config.py
```diff
@@ -0,0 +1,36 @@
+from dataclasses import dataclass
```
carolius, line 1: should probably add a multi gpu version of this as well

influence_benchmark/config/experiment_config.py
```diff
@@ -0,0 +1,154 @@
...
+    max_prompt_length: int
+    max_completion_length: int
+
+    def __post_init__(self):
+        super().__post_init__()
+        self.training_arg_keys = self.training_arg_keys + [
+            "beta",
+            "desirable_weight",
+            "undesirable_weight",
+            "max_length",  # TODO: How does this relate to the max_seq_length parameter
```
carolius, line 151: I think its the same, basically prompt_length + completion_length

influence_benchmark/experiments/hyperparam_sweeps/hyperparameter_sweep_kto.py
carolius, line 1: should also remove desirable_weight here i suppose

influence_benchmark/experiments/run_experiment.py
```diff
@@ -0,0 +1,73 @@
...
+    if torch.cuda.is_available():
+        print(f"Available CUDA devices: {torch.cuda.device_count()}")
+    else:
+        print("CUDA is not available.")
+
+    if config.seed is not None:
+        print(f"Setting all seeds to: {config.seed}")
+        set_all_seeds(config.seed)
+
+    mp.set_start_method("spawn", force=True)
```
carolius, line 37: is it fine to set this so late?

influence_benchmark/experiments/run_experiment.py
```diff
@@ -0,0 +1,73 @@
...
+    if isinstance(config, ExpertIterationConfig):
+        experiment_class = ExpertIteration
+        training_script_path = SFT_TRAINING_PATH
+    elif isinstance(config, KTOConfig):
+        experiment_class = KTO
+        training_script_path = KTO_TRAINING_PATH
+    else:
+        raise ValueError(f"Unknown experiment type: {type(config)}")
+
+    experiment = experiment_class(
```
carolius, line 52: this doesn't include override_initial_traj_path

carolius (commented):
There are a lot of type ignores which might be bad practice?
Also I can't figure out how the first lines are git-ignored? Nothing seems to suggest that?

influence_benchmark/RL/run_EI_iteration.py
micahcarroll: The thing I thought was nice about it is that this is immune to refactorings of the file name and location. What are the advantages of having it in `root.py`?

influence_benchmark/experiments/run_experiment.py
micahcarroll: Seems like it, at least when not on SLURM

influence_benchmark/experiments/run_experiment.py
micahcarroll: Whoops, my bad

influence_benchmark/config/experiment_configs/KTO_therapist_final.yaml
micahcarroll: I'm a bit confused about this – even if the ratio is 1.05, the scale of desirable and undesirable could be quite different?

micahcarroll: I guess are you saying we should just use `num_positives / num_negatives` for the undesirable weight and then use the ratio to figure out the desirable weight?

influence_benchmark/config/experiment_config.py
carolius: We could do, although it doesn't seem super important

influence_benchmark/config/experiment_configs/KTO_therapist_final.yaml
carolius: Yeah, only the ratio between them actually matters

influence_benchmark/RL/run_EI_iteration.py
carolius: I meant use the path library haha

influence_benchmark/utils/wandb_logging.py
```diff
@@ -155,15 +145,10 @@ def log_iteration_data_to_wandb(
...
-
-    if final_reward:
-        traj_df = group_turns_df_to_traj_df_final(turns_df)
-    else:
-        traj_df = group_turns_df_to_traj_df(turns_df)
-
     # This merge includes the traj-level reward/influence info to the turns_df entries for wandb logging.
     turns_df = turns_df.merge(traj_df, on=["env_name", "initial_state_id", "trajectory_id"])
+
+    # TODO: pretty sure there is some repeated computation in this function that we could easily avoid by using traj_df directly
```
AdhyyanNarang, line 151: I don't think there is repeated computation because the we merge the turns and traj dfs in the previous line, and then we access the traj info in the function.

But it is aesthetically displeasing; this is the one place where we are now we are storing traj level info in the turns df. It would be cleaner to pass both the turns and traj df to the function and extract the information there rather than combining across abstraction levels. But it's probably ok for now, not a huge deal

micahcarroll: Ah, i see, makes sense – renamed. I think this is fine honestly!

AdhyyanNarang:
Reviewed, looks good. I like that you added the "traj_" and "turns_" etc. prefixes to the variable names in wandb_logging.py and other files. 

This part of the code is definitely a lot cleaner now.

influence_benchmark/RL/EI.py
```diff
@@ -1,23 +1,16 @@
...
 
-    def _select_and_format_trajectories(self, trajectory_iteration_dir):
-        selected_trajectories, _ = get_best_worst_n_trajectories(
-            trajectory_iteration_dir, self.top_n_trajs_per_initial_state, final_reward=self.final_reward
-        )
-
-        self._format_and_save_trajectories_for_sft(selected_trajectories, trajectory_iteration_dir)
-
-    def _format_and_save_trajectories_for_sft(self, selected_partial_trajs, partial_traj_folder):
+    def _format_and_save_trajectories(self, selected_partial_trajs, partial_traj_folder):
```
AdhyyanNarang, line 10: What does the "partial" mean in this name? Is it because it contains the best/worst and we only need the best?

micahcarroll: It's technically not really entire trajectories, because each json will contain partial trajectories: trajectories up until a certain turn. Like if you take the length of this json, it will be a lot longer than the number of trajs (it's the number of timesteps)

influence_benchmark/config/accelerate_config.py
micahcarroll: Yeah, I guess do you have any that we can test? I don't think multi-gpu was working for me on RNN even without SLURM? I'll try again

influence_benchmark/experiments/run_experiment.py
```diff
@@ -0,0 +1,73 @@
...
+    parser = argparse.ArgumentParser(description="Expert Iteration Script")
+    parser.add_argument("--config", type=str, help="Path to the configuration file")
+    return parser.parse_args()
+
+
+def main():
+    args = parse_args()
+
+    config_path = args.config if args.config else DEFAULT_CONFIG_PATH
+    config = BaseExperimentConfig.load(config_path, devices=GPUS)
```
carolius, line 28: GPUS doesn't seem to be defined?

influence_benchmark/experiments/run_experiment.py
```diff
@@ -0,0 +1,73 @@
...
+from influence_benchmark.RL.KTO import KTO
+from influence_benchmark.RL.run_EI_iteration import SFT_TRAINING_PATH
+from influence_benchmark.RL.run_KTO_iteration import KTO_TRAINING_PATH
+from influence_benchmark.utils.utils import set_all_seeds
+
+# NOTE: specifying the GPUs here will override the ones in the config file
+
+
+def parse_args():
+    parser = argparse.ArgumentParser(description="Expert Iteration Script")
```
carolius, line 19: doesn't have to be EI does it?

influence_benchmark/config/experiment_config.py
```diff
@@ -0,0 +1,154 @@
...
+
+        if devices is not None:
+            print(f"Overriding GPUs from the config with GPU ids: {devices}")
+            config_dict["devices"] = devices
+
+        return cls.create_config(config_dict)
+
+    @classmethod
+    def create_config(cls: Type[T], config_dict: Dict[str, Any]) -> T:
+        cls._validate_config_keys(config_dict)
```
carolius, line 91: this runs before we set the config type which means that having a beta etc will cause a crash?

influence_benchmark/experiments/run_experiment.py
carolius: nor DEFAULT_CONFIG_PATH?

micahcarroll: Yes fixing this in next commits that I haven't pushed yet, this was a got ignore issue

influence_benchmark/config/experiment_config.py
micahcarroll: Ah, thanks for catching this 😬

influence_benchmark/RL/EI.py
AdhyyanNarang: I see, shouldn't the "_partial" be there in the base_iteration.py and KTO.py classes as well then? Looks like it's not there now

micahcarroll: Yeah maybe we should make the same change elsewhere

PR #90, opened 2024-08-17
Auto ignore_first_n_assistant_messages

micahcarroll:
This was very painful

### DIFF
--- a/influence_benchmark/RL/SFT.py
+++ b/influence_benchmark/RL/SFT.py
@@ -3,7 +3,7 @@ from dataclasses import dataclass, field
 from typing import Dict, Optional
 
 from accelerate import Accelerator
-from transformers import HfArgumentParser, TrainingArguments
+from transformers import AutoTokenizer, HfArgumentParser, TrainingArguments
 from trl import SFTTrainer
 
 from influence_benchmark.RL.conversation_collator import DataCollatorMaskingStaticConversation
@@ -45,6 +45,8 @@ def train_sft():
     if sft_config.seed is not None:
         set_all_seeds(sft_config.seed)
 
+    tokenizer = AutoTokenizer.from_pretrained(args.model_name)
+
     def format_dataset(example):
         r = {
             "text": tokenizer.apply_chat_template(example["messages"], tokenize=False),
@@ -52,7 +54,7 @@ def train_sft():
         }
         return r
 
-    dataset, model, tokenizer, peft_config = setup_dataset_and_model(args, format_dataset)
+    dataset, model, peft_config = setup_dataset_and_model(args, format_dataset, tokenizer)
 
     instruction_template = "<|start_header_id|>user<|end_header_id|>"
     response_template = "<|start_header_id|>assistant<|end_header_id|>"
--- a/influence_benchmark/RL/SFT.py
+++ b/influence_benchmark/RL/SFT.py
@@ -1,3 +1,4 @@
+# TODO: Rename this to sft-training or something
 from dataclasses import dataclass, field
 from typing import Dict, Optional
 
@@ -35,7 +36,6 @@ class ScriptArguments:
     lora_dropout: Optional[float] = field(default=None)
     max_seq_length: Optional[int] = field(default=None)
     g_c_kwargs: Dict = field(default_factory=lambda: {"use_reentrant": False})
-    ignore_first_n_assistant_messages: int = field(default=0)
     lora_path: Optional[str] = field(default=None)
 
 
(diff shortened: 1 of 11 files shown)


influence_benchmark/RL/conversation_collator.py
```diff
@@ -106,30 +104,32 @@ def torch_call(self, examples: List[Union[List[int], Any, Dict[str, Any]]]) -> D
...
                 for human_start, response_start in zip(
-                    human_token_ids_idxs[self.ignore_first_n_messages :],
-                    response_token_ids_idxs[self.ignore_first_n_messages :],
+                    human_token_ids_idxs[num_messages_to_ignore:],
+                    response_token_ids_idxs[num_messages_to_ignore:],
                 ):
                     batch["labels"][i, human_start:response_start] = self.ignore_index
-            else:
-                for response_start in response_token_ids_idxs[self.ignore_first_n_messages :]:
+            else:  # TODO: what is the alternative to instruction template and why are we overriding start_idx?
```
micahcarroll, line 126: Flagging this

micahcarroll:
Diffs will be smaller once https://github.com/carolius/Influence-benchmark/pull/87 is merged

influence_benchmark/RL/conversation_collator.py
```diff
@@ -106,30 +104,32 @@ def torch_call(self, examples: List[Union[List[int], Any, Dict[str, Any]]]) -> D
...
+                    start_idx = len(batch["labels"][i])  # Ignore all if not enough messages
+            else:
+                if len(response_token_ids_idxs) > num_messages_to_ignore:
+                    start_idx = response_token_ids_idxs[num_messages_to_ignore - 1]
                 else:
-                    if len(response_token_ids_idxs) > self.ignore_first_n_messages:
-                        start_idx = response_token_ids_idxs[self.ignore_first_n_messages - 1]
-                    else:
-                        start_idx = len(batch["labels"][i])  # Ignore all if not enough messages
+                    start_idx = len(batch["labels"][i])  # Ignore all if not enough messages
```
micahcarroll, line 117: @carolius I want to chat more with you about this – note to self: when predicting tokens for "gap 2", can the system see the tokens for "gap 1"?

influence_benchmark/RL/SFT.py
```diff
@@ -121,6 +123,8 @@ def formatting_prompts_func(example):
         data_collator=collator,
         max_seq_length=args.max_seq_length,
     )
+    # Remove the columns that are not needed or it will cause errors, as training will try to cast these strings to tensors
+    trainer.train_dataset = trainer.train_dataset.remove_columns(["text", "messages"])
```
carolius, line 87: I'm a bit confused about why this would show up now?

carolius (commented):
I think it's quite hard to tell if the ignoring is happening correctly, have you checked a few examples in the debugger to see if the right stuff is being masked?

micahcarroll:
> I think it's quite hard to tell if the ignoring is happening correctly, have you checked a few examples in the debugger to see if the right stuff is being masked?

I tried checking that we were getting the right `num_messages_to_ignore` for each traj by creating a fake traj file that had different "hardcoded_msgs" nums that I set myself. Plus, the test run didn't seem to be affected in any meaningful way by this change. 

That being said, I didn't check the masking out logic carefully. It was kinda complicated, and my changes didn't actually touch it as far as I could tell. I'm just giving it a different argument for each item in the batch.

influence_benchmark/RL/expert_iteration.py
```diff
@@ -14,13 +14,23 @@ def _select_and_format_trajectories(self, trajectory_iteration_dir):
...
+        for partial_traj in selected_partial_trajs:
+            messages_so_far = self.format_valid_messages(partial_traj)
 
-            formatted_trajectories.append({"messages": messages})
+            roles = set(msg["role"] for msg in messages_so_far)
+            assert roles == {
+                "system",
+                "user",
+                "assistant",
+            }, "Other roles may mess up the calculation that follows, make sure the code still works and remove this assertion."
```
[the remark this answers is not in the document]
micahcarroll: Yeah, I was just punting it to later 😅

influence_benchmark/RL/SFT.py
micahcarroll: In order to access the `num_messages_to_ignore` for each training example at Collator time, I needed to add the `remove_unused_columns=False` arg to SFT, which by default causes the errors described in this comment

micahcarroll: (Figuring out how to do this was the painful part)

influence_benchmark/RL/conversation_collator.py
micahcarroll, line 126:
```suggestion
            else:  # If there is no system prompt, we want to train on all assistant messages after start_idx but not before, so we need to figure out where that is
```

PR #91, opened 2024-08-17
Fixing bug in computing average rewards across trajectories

AdhyyanNarang:
Earlier there was a bug with reporting top N reward, where it could be smaller than the overall average reward. This bug is now resolved and the entire code using pandas is rewritten to be more readable to prevent any future such bugs.

- The plot_preferences_per_iteration.py file does not suffer from the bug anymore.
- The expert iteration code chooses the same trajectories as before, and the function definition for get_best_worst_n_trajectories is unchanged.


.gitignore
```diff
@@ -5,4 +5,5 @@ data/*
 wandb
 *.out
 **/tmp/**
-scratch.ipynb
\ No newline at end of file
+scratch.ipynb
+**/*dummy.ipynb
```
micahcarroll, line 9: Maybe just rename yours to `scratch.ipynb` and we'll be good
```suggestion
```

influence_benchmark/stats/preferences_per_iteration.py
```diff
@@ -62,57 +33,33 @@ def get_best_worst_n_trajectories(
...
+
     # Load all trajectories from files
-    traj_timestep_df = load_trajectories(trajectory_path)
+    turns_df = load_turns_df_from_traj_path(trajectory_path)
+
     if final_reward:
         rew_label = "traj_final_rew"
-        infl_label = "traj_final_infl"
-        rewards_df = compute_final_turn_rewards(traj_timestep_df)
+        # infl_label = "traj_final_infl"
```
micahcarroll, line 42: Maybe we can clean this up?  Are we not computing influence anymore?

influence_benchmark/stats/preferences_per_iteration.py
```diff
@@ -62,57 +33,33 @@ def get_best_worst_n_trajectories(
...
-        .apply(
-            lambda x: x.assign(
-                n_trajectories=len(x),
-                avg_rew_across_trajs_with_init_s=x[rew_label].mean(),
-                avg_infl_across_trajs_with_init_s=x[infl_label].mean(),
-            ).pipe(func, n_chosen_trajs, rew_label)
-        )
-        .reset_index(drop=True)
-    )
+        # infl_label = "traj_mean_infl"
```
micahcarroll, line 46: Same here

influence_benchmark/stats/utils_pandas.py
```diff
@@ -0,0 +1,102 @@
...
+        ]
+        .mean()
+        .reset_index()
+        .rename(columns={"timestep_reward": "traj_mean_rew", "timestep_influence_level": "traj_mean_infl"})
+    )
+    return traj_df
+
+
+def get_filtered_turns_df(turns_df, filtered_traj_df):
+    return pd.merge(turns_df, filtered_traj_df, on=["env_name", "initial_state_id", "trajectory_id"])
```
micahcarroll, line 64: Maybe add some documentation here? What does this do exactly? What kinds of inputs is it expecting?

influence_benchmark/utils/wandb_logging.py
```diff
@@ -155,15 +155,15 @@ def log_iteration_data_to_wandb(
...
     if final_reward:
-        rew_df = compute_final_turn_rewards(traj_timesteps_df)
+        traj_df = group_turns_df_to_traj_df_final(turns_df)
     else:
-        rew_df = compute_average_traj_rewards(traj_timesteps_df)
+        traj_df = group_turns_df_to_traj_df(turns_df)
 
-    traj_timesteps_df = traj_timesteps_df.merge(rew_df, on=["env_name", "initial_state_id", "trajectory_id"])
-    trajectories, env_stats = extract_wandb_data(traj_timesteps_df)
+    turns_df = turns_df.merge(traj_df, on=["env_name", "initial_state_id", "trajectory_id"])
```
micahcarroll, line 165: Maybe add a comment for this line – what is this merge achieving exactly?

micahcarroll: I guess adding the reward/influence info to the turns_df traj?

influence_benchmark/stats/utils_pandas.py
```diff
@@ -0,0 +1,102 @@
...
+    # Average over turns, will include num_envs * num_initial_states * num_trajs_per_initial_state rows
+    traj_df = (
+        turns_df.groupby(["env_name", "initial_state_id", "trajectory_id"])[
+            ["timestep_reward", "timestep_influence_level"]
+        ]
+        .mean()
+        .reset_index()
+        .rename(columns={"timestep_reward": "traj_mean_rew", "timestep_influence_level": "traj_mean_infl"})
+    )
+    return traj_df
```
micahcarroll, line 60: I feel like the names for these functions are a bit uninformative. Seems like the point here is to add reward/influence info, but that isn't part of the name at all

micahcarroll: for doing this kinds of renaming tasks, you'll probably find fn+F2 to be really helpful after clicking on the method name

influence_benchmark/stats/utils_pandas.py
```diff
@@ -0,0 +1,102 @@
...
+    )
+
+    # Calculate average reward and influence across top_n trajectories
+    top_n_avg = (
+        filtered_traj_df.groupby(["env_name", "initial_state_id"])
+        .agg(avg_rew_top_n_trajs=("traj_mean_rew", "mean"), avg_infl_top_n_trajs=("traj_mean_infl", "mean"))
+        .reset_index()
+    )
+
+    # Step 3: Merge the two DataFrames to create state_df
```
micahcarroll, line 100: Nit: there is no step 1/2 :)

influence_benchmark/stats/utils_pandas.py
```diff
@@ -0,0 +1,102 @@
...
+            lambda x: x.assign(
+                n_trajectories=len(x),
+            ).pipe(func, num_chosen_trajs, rew_label)
+        )
+        .reset_index(drop=True)
+    )
+    return filtered_df
+
+
+def group_traj_df_to_state_df(traj_df, filtered_traj_df):
```
micahcarroll, line 81: Why is this a "state_df", while the others are traj_dfs? I remember you mentioning this, but probably worth adding to a comment because I don't think it's currently self-evident

micahcarroll (commented):
Looks good apart from minor things! Sorry for leaving so many nit-style comments, just think that this part of the code is both very important and hard to parse because of pandas. So pretty important to leave signposts for our future selves if we ever have to touch it again (we will for excluding train/test trajs)

influence_benchmark/stats/utils_pandas.py
AdhyyanNarang: Prior to this, we do the filtering at the level of trajectories to choose the best/worst ones.

This line extracts the relevant turns from turns_df that correspond to these filtered trajs for training. Will add a comment.

influence_benchmark/stats/utils_pandas.py
AdhyyanNarang: Renamed all "state" to subenv

influence_benchmark/utils/wandb_logging.py
AdhyyanNarang: Yep that's right. Added comment

PR #92, opened 2024-08-18
Fixing 0-loss-issue

micahcarroll:
- Made the logic in the collator a lot simpler and less flexible (because I don't think we actually were using or planning to use the additional flexibility anywhere)

Happy to incorporate other kinds of roles in the middle of the conversation, but need to know exactly how they are being used


influence_benchmark/RL/conversation_collator.py
```diff
@@ -67,69 +67,60 @@ def __init__(
...
+
+                # self.user_template -> '<|start_header_id|>user<|end_header_id|>'
+                user_token_slice = batch["labels"][i][idx : idx + len(self.user_template_ids)].tolist()
+                if self.user_template_ids == user_token_slice:
+                    user_template_start_idxs.append(idx)
+
+            num_assistant_msgs = len(assistant_response_start_idxs)
+            num_user_msgs = len(user_template_start_idxs)
+
+            assert num_assistant_msgs == num_user_msgs, "The number of assistant and user messages should be the same"
```
carolius, line 103: this might no longer be true when we add tool use depending on how we do it but we can fix it later i suppose

micahcarroll: I'm also happy to fix it at that time. But we should definitely revisit this code once we make those changes

PR #105, opened 2024-08-22
Add state percentages and conversation length logging

carolius:
## Description
Added wandb logging for the percentage of visited states and the average conversation length.

## Type of change

## How Has This Been Tested?


micahcarroll:
I believe I've resolved some of the merge conflicts from Ady's branch

influence_benchmark/utils/wandb_logging.py
```diff
@@ -176,15 +201,20 @@ def log_iteration_data_to_wandb(turns_df, traj_df, iteration_step, top_n_trajs_p
             f"Avg influence: {env_avg_infl:.2f} ({env_stderr_infl:.2f})"
```
micahcarroll, line 201: I think it may be nice to also include the percentage stats here? Would require moving this whole print statement after the stats are computed

micahcarroll:
This currently doesn't pass the pytest but will once #110 is merged into it

PR #107, opened 2024-08-22
Log stderr on wandb and deal with invalid probs

AdhyyanNarang:
## Description
Log stderr on wandb and deal with invalid generated probs

## Type of change

- Include stderr on logging
- Include stderr on wandb
- For all accessor models, check whether the generated probabilities are in the simplex.
- In cases where they are not, log a warning message.
- In cases where probs are fixable, using normalization to fix

## How Has This Been Tested?

- Ran local tests for new function and using EI_test.yaml


carolius (commented):
I'm a bit worried this will conflict with the other PRs, I haven't checked but I implemented the length penalty as a -1 preference with a certain weight, explicitly making preference probs not sum to one. It might just be fine but we should make sure nothing weird happens

influence_benchmark/stats/preferences_per_iteration.py
```diff
@@ -65,10 +66,27 @@ def compute_iteration_statistics(traj_df: pd.DataFrame, top_n: int) -> Dict[str,
     traj_df_filtered = filter_traj_df(traj_df, num_chosen_trajs=top_n, func=pd.DataFrame.nlargest)
     subenv_df = group_traj_df_to_subenv_df(traj_df, traj_df_filtered)
 
+    num_trajs = len(traj_df["traj_rew"])
+    num_filtered_trajs = len(traj_df_filtered["traj_rew"])
+
     results["rew_avg_all_trajs"] = subenv_df["mean_traj_reward"].mean()
+    results["rew_stderr_all_trajs"] = (np.std(traj_df["traj_rew"], ddof=1) / np.sqrt(num_trajs)) if num_trajs > 1 else 0
```
micahcarroll, line 73: I think we should probably have a util for this. I usually write a `mean_and_std_err` util in other projects

influence_benchmark/utils/wandb_logging.py
```diff
@@ -160,14 +160,33 @@ def log_iteration_data_to_wandb(turns_df, traj_df, iteration_step, top_n_trajs_p
...
+
+        reward_array = stats["traj_reward_n"]
+        influence_array = stats["traj_influence_n"]
+        num_traj = len(reward_array)
+
+        env_avg_rew = np.mean(reward_array)
+        if num_traj > 1:
+            env_stderr_rew = np.std(reward_array, ddof=1) / np.sqrt(num_traj)
+        else:
+            env_stderr_rew = 0
```
micahcarroll, line 172: I would do this with a one liner once the util is implemented

influence_benchmark/stats/preferences_per_iteration.py
micahcarroll: Also would we ever pass num_trajs = 0? Maybe we should just have an assert that it's non-zero?

micahcarroll (commented):
Mostly looks good, but I think would be good to change a couple of these small things I commented on

influence_benchmark/stats/preferences_per_iteration.py
AdhyyanNarang: Don't think we'd ever pass empty array but the computation fails when num_traj = 1 cuz of the ddof=1. This is cuz it's the sample std, which needs the array to have multiple elements.

micahcarroll:
Probably worth merging dev into this branch and running the tests again, because EI_test.yaml was not actually running the wandb code up until now

PR #108, opened 2024-08-23
Improved config system to be able to do small deltas from other configs

micahcarroll:
## Description
Now configs can inherit from other ones. Now we don't have to updates 20 configs every time if we change a key, and we can be sure that all params are the same except ones explicitly specified.

## Type of change

## How Has This Been Tested?

- [ ] I am running two large scale exps, with the new two configs I added


carolius:
Looks good

PR #109, opened 2024-08-23
Across iteration LR decay, improved autocopy_and_sbatch

micahcarroll:
## Description

- Added param for having LR decay across iterations
- Made the `sbatch` parent scripts much more minimal
- Now `autocopy_and_sbatch` auto uses all visible gpus when running `run_experiment.py` (this solves a mistake that was easy to make if run_experiments.py had GPUs defined at the time that one was using an sbatch script). Now the global variables in `run_experiment.py` are entirely overridden by all `sbatch` scripts
- Improved SLURM naming when running `sbatch` (now the same as the config file that's being run)
- Improved SLURM logging filename to be more easily identifiable (config name + timestamp)
- *After this PR, we should probably commit all the configs we're using before we run any experiment*

## Type of change

## How Has This Been Tested?

- [x] I have run one or more full experiment(s): results look good, and autocopy worked fine

I've tested things pretty extensively I believe this time around. Trying to prevent breaks like the ones we caused previously


carolius:
Does this mean that we no longer specify what experiment to run in run_experiment?

micahcarroll:
You can if you want to run it locally with the play button in vscode 

But if you're trying to run something on slurm, you should just specify the config file in kto.sh (it will ignore whatever is written in run_experiment.py)

micahcarroll:
Also the gpus when kicking off SLURM runs is always set to be all visible devices (regardless of what run_experiment.py says)

influence_benchmark/config/experiment_config.py
```diff
@@ -143,7 +143,8 @@ class LocalTrainingConfig(BaseExperimentConfig):
     report_to: str
     optim: str
     max_length: int
-    lr_scheduler_type: str
+    lr_scheduler_type: str  # (Within each iteration)
+    across_iter_lr_decay_rate: float  # (Across iterations) E.g. if 0.5, LR will be halved after every iteration
```
carolius, line 147: Probably not be worth changing but it might be a bit confusing to call it decay rate? if i understand it correctly 1 means no decay so this is like a not decay rate. Also your example is the only possible one where it is not clear which it means since both a decay rate of 0.5 and a not decay rate of 0.5 are the same

influence_benchmark/RL/EI.py
```diff
@@ -7,8 +7,8 @@ class ExpertIteration(BaseIteration):
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
 
-    def _format_and_save_trajectories(self, selected_partial_trajs, partial_traj_folder):
-        best_trajectories, _ = selected_partial_trajs
+    def _format_and_save_trajectories(self, selected_trajectories, trajectory_folder):
```
carolius, line 10: There is a lot of going back and forth with variable names, IIRC you were the one that changed the variable name to selected_trajectories, then changed it to selected_partial_trajs and are now changing it back to selected_trajectories 😂. selected_trajectories is probably the best one out of these though

influence_benchmark/RL/run_EI_iteration.py
```diff
@@ -48,7 +49,10 @@ def train_sft():
     sft_config.remove_unused_columns = False  # Necessary for the collator to have access to traj metadata
     sft_config.gradient_checkpointing_kwargs = args.g_c_kwargs
     sft_config.dataset_text_field = "text"
-
+    sft_config.learning_rate = sft_config.learning_rate * (args.across_iter_lr_decay_rate**args.iteration)
```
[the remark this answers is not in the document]
micahcarroll: yeah, it always has to be defined in the config file. In all config files except the ones that I tested, I set it to 1.0

influence_benchmark/config/experiment_config.py
micahcarroll: loool fair

influence_benchmark/RL/EI.py
micahcarroll: I know, Ady pointed out it was a bad name, and it wasn't changed everywhere, so I switched it back to this. And in the parts of the code that are only used with `top` trajs, I used that

influence_benchmark/RL/run_EI_iteration.py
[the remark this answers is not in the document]
carolius: Maybe we shouldn't have the type be optional then?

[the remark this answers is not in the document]
micahcarroll: Currently I don't have anything be optional in configs. In my experience, optional vars often lead to silent failures. This is also an issue with config inheritance, but to a lesser extent. Everything being explicit is safer (even though it's a bit more annoying)

PR #115, opened 2024-08-30
Retroactive evals

AdhyyanNarang:
## Description
Retroactive evals of influence related metrics

## Type of change

- New script for retroactive evals. Uses the AssessorModel instead of rewriting the code to keep it clean
- Multiprocessing for Llama
- Support for GPT with async processing
- Helper file and notebook for plotting and exploring different runs

## How Has This Been Tested?
Ran the script on KTO 1-step data, and verified that curves with the PM prompt look similar to wandb


micahcarroll:
Running this script on [this run](https://wandb.ai/influence_benchmark/influence-benchmark/runs/tntth1yc/overview) (I've sent you the trajs over Slack), and having a simple plot of how these metrics change over time could be a good thing to look at.

Also you could try add the PM prompt that was used for the run into your _master_config, and see if it tracks with the reward that is on the wandb

carolius:
It might make sense to add support for the openai backend while you're at it, I think it could be good to use a different model than llama-8b to measure the influence

influence_benchmark/stats/retroactive_evals.py
```diff
@@ -0,0 +1,172 @@
...
+        self.batch_size = batch_size
+        self.devices = [f"cuda:{i}" for i in devices]
+
+        if (self.batch_size % len(self.devices)) != 0:
+            self.batch_size = (self.batch_size // len(self.devices)) * len(self.devices)
+
+        assert self.batch_size % len(self.devices) == 0
+
+        self.assessor_models = {metric: AssessorModel(config[metric]) for metric in metrics}
+        self.state_variables = {"agent_name": "Agent", "User": "User"}
```
carolius, line 55: shouldn't it be getting User from env_char_label? maybe it's bad to define it it two places?

influence_benchmark/stats/retroactive_evals.py
```diff
@@ -0,0 +1,172 @@
...
+
+    def evaluate_transcript(self, transcript, backend):
+        # Evaluate the transcript for each metric using the provided backend
+        return {metric: self.get_eval_from_backend(transcript, metric, backend) for metric in self.metrics}
+
+    def get_eval_from_backend(self, transcript, metric, backend):
+        state = RetroactiveIterationState(history=transcript, variables=self.state_variables)
+        messages = self.assessor_models[metric].prepare_messages(state)
+        valid_tokens = self.config[metric]["valid_tokens"]
+        responses = backend.get_next_token_probs_normalized_vec([messages], valid_tokens_n=[valid_tokens])
```
carolius, line 135: might be cleaner if you go via vectorized assesor model, I think that's where you put your normalization code as well. Probably not that important

AdhyyanNarang: I see, yes. For now, I think it is OK because it will not change the functionality too much, but would be happy to make this change later if we want. It may interact in intricate ways with the mp/async code I set up right now so that's why I feel like it's not worth a ton of time. 

For the normalization, I moved those functions to a utils file so I can call them here too

influence_benchmark/stats/retroactive_evals.py
AdhyyanNarang: Fixed

influence_benchmark/stats/retroactive_evals.py
```diff
@@ -0,0 +1,251 @@
...
+from influence_benchmark.root import LOADED_DOTENV
+from influence_benchmark.stats.utils_pandas import calculate_expectation, load_turns_df_from_iteration_path
+from influence_benchmark.utils.utils import load_yaml, model_name_to_backend_class
+from influence_benchmark.utils.utils_prob import check_simplex_and_transform
+
+
+@dataclass
+class RetroactiveIterationState:
+    history: List[Dict[str, str]]
+    variables: Dict[str, str]
```
micahcarroll, line 22: This is a bit hacky as the AssessorModel is built to deal with State instances from state.py. But it's probably fine. Maybe worth adding a comment about that

influence_benchmark/stats/retroactive_evals.py
```diff
@@ -0,0 +1,251 @@
...
+        variables = {
+            "agent_name": "Agent",
+            "user_name": self.config[metric]["env_char_label"],
+        }
+        if metric == "preference":
+            variables["pm_prompt"] = pm_prompt
+
+        state = RetroactiveIterationState(history=transcript, variables=variables)
+        messages = self.assessor_models[metric].prepare_messages(state)
+        valid_tokens = self.config[metric]["valid_tokens"]
```
micahcarroll, line 149: A bunch of this code seems repeated between the sync and async modes for GPT and llama

micahcarroll: Also seems like there is a lot of repetition from get_response for VectorizedAssessorModel

influence_benchmark/stats/retroactive_evals.py
```diff
@@ -0,0 +1,251 @@
...
+
+        responses = backend.get_next_token_probs_normalized_vec([messages], valid_tokens_n=[valid_tokens])
+
+        # Use check_simplex_and_transform to handle invalid probabilities
+        _, outputs_fixed = check_simplex_and_transform(prob_dict=responses[0], log_name=f"{metric}_eval")
```
micahcarroll, line 242: Seems like there is a lot of repetition from get_response for VectorizedAssessorModel, maybe worth using that directly?

micahcarroll (changes requested):
Thanks @AdhyyanNarang for the changes! I think there are still some things that can be made cleaner and without repetition. That will be helpful in reducing cost if we ever want to modify other parts of the code.

Did a bit of cleaning myself (imports, removing defaults, type checker stuff), but I think there is a bit of restructuring worth doing before we should merge

influence_benchmark/stats/plot_retroactive_evals.py
```diff
@@ -0,0 +1,256 @@
...
+    eval_config = load_yaml(str(PROJECT_ROOT / config["eval_config"]))
+    backend_config = config["backend_config"]
+    run_name = config["run_name"]
+    metrics = config["metrics"]
+    batch_size = config["batch_size"]
+    gpu_ids = config["gpu_ids"]
+    iteration_number = config["iteration_number"]
+    save = config["save"]
+    env_name_prefix = config.get("env_name_prefix", "")
+    env_config_path = PROJECT_ROOT / config.get("env_config_path", "config/env_configs/normal_therapist")
```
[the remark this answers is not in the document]
micahcarroll: We currently have these lines repeated also in the notebook 2 times I think

influence_benchmark/stats/retroactive_evals.py
```diff
@@ -0,0 +1,256 @@
...
+
+        return self._process_results(results)
+
+    async def _async_evaluate_batch(self, batch, start_index):
+        async with self.semaphore:
+            process_backend = self.backend_class(
+                model_name=self.backend_config["model_name"],
+                model_id=self.backend_config["model_id"],
+                lora_path=self.backend_config["lora_path"],
+                device=None,  # type: ignore
```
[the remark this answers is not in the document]
AdhyyanNarang: Any of these async functions are called if using the GPTBackend. For the HF backend, we do a multiprocessing approach across devices in the sync functions. Added comments to make this clearer.

influence_benchmark/config/experiment_configs/retroactive_evals/retro_single.yaml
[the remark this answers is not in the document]
AdhyyanNarang: I think it should be OK either way. I thought having the config system was cleaner to separate concerns but it's not a strong preference, so I went ahead and removed the configs

influence_benchmark/stats/plot_retroactive_evals.py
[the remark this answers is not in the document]
AdhyyanNarang: I think for transparency it's better to have the constructor of `RetroactiveIterationEvaluator` spell out exactly the inputs it is expecting rather than taking in a dict. But now that we do not have a config, I just set these parameters in python in the script/notebook so it should be fine.

It was repeated in the notebook two times because it was loading 2 different config files -- one for a single iteration and the other for all iterations in a run. But now that is set in python too, so it's not repeated.

influence_benchmark/config/experiment_configs/retroactive_evals/retro_single.yaml
```diff
@@ -0,0 +1,16 @@
...
+run_name: kto-mixed-therapist-1-step-09-04_14-47
+metrics:
+  - preference
+batch_size: 16
+gpu_ids:
+  - 4
+iteration_number: 15
+save: true
+env_name_prefix: normal
+env_config_path: config/env_configs/normal_therapist
```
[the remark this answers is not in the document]
AdhyyanNarang: I think this may be helpful for the cross-env generalization, if we want to evaluate on different environments. So I propose we leave it for now unless you feel strongly otherwise. If I don't end up using it there, I can delete these in that PR

influence_benchmark/stats/retroactive_evals.py
```diff
@@ -0,0 +1,251 @@
...
+        return pm_prompts
+
+    def evaluate_iteration(self) -> pd.DataFrame:
+        # Extract all transcripts and env_names from the trajectory DataFrame
+        all_transcripts_with_env = list(zip(self.traj_df["history"].tolist(), self.traj_df["env_name"].tolist()))
+
+        if self.env_name_prefix:  # Changed this condition
+            all_transcripts_with_env = [
+                (transcript, f"{self.env_name_prefix}_{env_name}") for transcript, env_name in all_transcripts_with_env
+            ]
```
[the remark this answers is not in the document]
AdhyyanNarang: I had it because was testing on the KTO 1-step run. I removed the hardcoded "_" so that if using newer runs, we could set the prefix to be "" and it should be compatible with everything

influence_benchmark/config/experiment_configs/retroactive_evals/retro_single.yaml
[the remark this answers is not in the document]
micahcarroll: Ah I see, makes sense!

influence_benchmark/stats/retroactive_evals.py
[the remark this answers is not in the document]
micahcarroll: I see, yeah I think it's fine if we're not backward compatible!

influence_benchmark/stats/retroactive_evals.py
AdhyyanNarang: This would be possible for HF but not for GPT because the VectorizedAccessorModel does not have async methods; so I think it is better to stick with AccessorModel to keep things consistent.

But I made some helper instance methods that should reduce the reuse and increase readability a good amount.

influence_benchmark/stats/retroactive_evals.py
AdhyyanNarang: Added some helper functions to make this cleaner

influence_benchmark/stats/retroactive_evals.py
carolius: I'm confused by what you mean? The VectorizedAccessorModel uses GPTBackend which does use async? If you use get_next_token_probs_normalized_vec (which VectorizedAccessorModel does) it will run the whole batch at the same time with async

micahcarroll: I'll wait on this possibility to be explored before reviewing further changes!

AdhyyanNarang: Ok yes I see what you mean. Have updated the file accordingly and committed changes.

carolius (commented):
LGTM

micahcarroll:
I guess this PR still doesn't pull out a TrajectoryGeneratorClass?

AdhyyanNarang:
> I guess this PR still doesn't pull out a TrajectoryGeneratorClass? That may be a nice thing to add!

Wasn't that discussion for the cross generalization PR? Seems more relevant for that one since this one is more concerned with evaluating already-generated trajs on new metrics

micahcarroll:
Oh sorry, yes, makes sense!

influence_benchmark/stats/retroactive_evals.py
```diff
@@ -0,0 +1,266 @@
...
+        flat_results = list(itertools.chain(*results))
+        return self.process_results(flat_results)
+
+    def evaluate_batch(self, batch, start_index, device):
+        backend = self.backend_class(
+            model_name=self.backend_config["model_name"],
+            model_id=self.backend_config["model_id"],
+            lora_path=self.backend_config["lora_path"],
+            device=device,
+        )
```
micahcarroll, line 218: This step takes some time to run – seems strange to have it repeated it every batch. Is there maybe a way to give multiple batches to a process and have the process go through multiple batches, so that we don't have to reload the backend a bunch of times?

influence_benchmark/stats/retroactive_evals.py
```diff
@@ -0,0 +1,266 @@
...
+            config (Dict): Configuration for the evaluator.
+            metrics (List[str]): List of metrics to evaluate.
+            batch_size (int): Batch size for processing.
+            devices (List[str]): List of GPU devices to use.
+            env_name_prefix (str): Prefix for environment names.
+            env_config_path (Path): Path to environment configuration files for preference prompts.
+        """
+        self.turns_df = load_turns_df_from_iteration_path(iteration_path)
+
+        self.traj_df = self.turns_df.loc[self.turns_df.groupby("trajectory_id")["turn"].idxmax()]
```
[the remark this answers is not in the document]
AdhyyanNarang: Makes sense, this is better

micahcarroll:
I'm sorry, I made a big mess in trying to fix some issues with this branch. Now this PR is subsumed by #129 (it has all the changes from this branch, plus some changes I made). If you could address the new comments I left above (plus some I'll leave in #129), branching off from #129, that would be great! 🙏

influence_benchmark/stats/retroactive_evals.py
AdhyyanNarang: I see. The simplest way I can think of to avoid this issue is to just have a self.backends_lst in the constructor, with one element for each device. Then we can just load the backends once and pull it out in this function

PR #120, opened 2024-09-03
Dotenv hotfix

micahcarroll:
## Description
Fixes the issue with SLURM and `.env`: previously `~/.env` was not visible from SLURM, so I have moved it in the project directory itself and gitignored it.

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)

## How Has This Been Tested?

- [x] I have tested with SLRUM that now it works

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,3 +9,4 @@ scratch.ipynb
 scratch.yaml
 n_test
 n_test1
+.env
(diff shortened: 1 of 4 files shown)


micahcarroll:
Actually still seems to not work

carolius:
Seems to work

PR #122, opened 2024-09-03
Sample top reward and bottom reward trajs to log to wandb

weissercn:
## Description
Sample top reward and bottom reward trajs to log to wandb

Addressing https://www.notion.so/micahcarroll/Sample-top-reward-and-bottom-reward-trajs-to-log-to-wandb-07418f21ac11448a97b332ea2f9e27a9?pvs=4

## Type of change

## How Has This Been Tested?

## Example functionality

For the following settings (chosen to discover edge cases like having multiple turns):
Screenshot 2024-09-03 at 7 36 27 PM
The original turns dataframe looks like this. Previously, we randomly sampled from this dataframe for logging
image

Here are the new dataframes with 3 top and 1 bottom trajectories for each environment
image

This resulted in the following wandb
https://wandb.ai/influence_benchmark/influence-benchmark/runs/4aav2ela?nw=nwuserweissercn

## Details

For each environment take the 3 top trajectories (after deduplicating trajectories on the level of turns)


influence_benchmark/utils/wandb_logging.py
```diff
@@ -205,10 +206,35 @@ def print_stats_and_log_to_wandb(turns_df, traj_df, iteration_step, top_n, trajs
...
+        top_trajectories = get_trajs_wandb_html(top_turns_df)
+
+        for trajectory in top_trajectories:
+            wandb.log(
+                {
+                    f"Iteration {iteration_step}, Env: {trajectory['env_name']} HighestRewTraj": wandb.Html(
+                        trajectory["html_content"]
+                    )
+                }
+            )
```
micahcarroll, line 224:
```suggestion
```

influence_benchmark/utils/wandb_logging.py
```diff
@@ -205,10 +206,35 @@ def print_stats_and_log_to_wandb(turns_df, traj_df, iteration_step, top_n, trajs
...
+
+        bottom_trajectories = get_trajs_wandb_html(bottom_turns_df)
+        for trajectory in bottom_trajectories:
             wandb.log(
-                {f"Iteration {iteration_step}, Env: {trajectory['env_name']}": wandb.Html(trajectory["html_content"])}
+                {
+                    f"Iteration {iteration_step}, Env: {trajectory['env_name']} LowestRewTraj": wandb.Html(
+                        trajectory["html_content"]
+                    )
+                }
```
micahcarroll, line 239:
```suggestion
                {f"Iteration {iteration_step}, Env: {trajectory['env_name']}": wandb.Html(trajectory["html_content"])}
```

influence_benchmark/utils/wandb_logging.py
```diff
@@ -205,10 +206,35 @@ def print_stats_and_log_to_wandb(turns_df, traj_df, iteration_step, top_n, trajs
...
+            )
+
+        bottom_turns_df: pd.DataFrame = get_logging_turns_df(
+            turns_df,
+            num_chosen_trajs=bottom_n_to_log,
+            func=pd.DataFrame.nsmallest,
+        )
+
+        bottom_trajectories = get_trajs_wandb_html(bottom_turns_df)
+        for trajectory in bottom_trajectories:
```
micahcarroll, line 233:
```suggestion
        for trajectory in bottom_trajectories + top_trajectories:
```

influence_benchmark/utils/wandb_logging.py
micahcarroll: I think I prefer this way of logging that puts the bottom and top ones together in the same wandb panel. We already have far too many wandb panels

weissercn:
An alternative solution that seems more error-prone and duplicated is
1. Select top-k trajectories for each environment in traj_df
2. Then for each column in the filtered traj_df find the list of matching columns in turns_df (using a join or a lookup)
3. Out of this list, select the one with the highest turn. 
4. Join this result back to the traj_df dataframe



The current implementation seems simpler and only needs turn_df, rather than turn_df and traj_df and has simpler steps
1. Deduplicate turn_df based on turn number (pick the highest turn)
2. For each environment pick the top-k rows of this dataframe

What do you think @micahcarroll ? I am happy to implement the other version if you prefer that

influence_benchmark/stats/utils_pandas.py
```diff
@@ -118,6 +118,35 @@ def get_selected_turns_df(turns_df: pd.DataFrame, selected_traj_df: pd.DataFrame
...
+    Returns:
+        pd.DataFrame: DataFrame containing the top-k trajectories for each environment with all of the same columns as the original dataframe.
+    """
+    # Group by env_name, initial_state_id, and trajectory_id, then get the row with max turn
+    df_max_turn = turns_df.loc[turns_df.groupby(["env_name", "initial_state_id", "trajectory_id"])["turn"].idxmax()]
+
+    # Now group by env_name and initial_state_id, and get the top k based on traj_rew
+    top_k_df = (
+        df_max_turn.groupby(groupby_feats).apply(lambda x: func(x, num_chosen_trajs, "traj_rew")).reset_index(drop=True)
+    )
```
micahcarroll, line 145: For an experiment with `max_turns=2`, say that `len(turns_df)=2` (with 1 trajs)

Would `len(top_k_df) = 2` in this case, or `len(top_k_df)=1`? I think it would be len 1, but if that's the case, I think it may break the trajectory logging, because `get_trajs_wandb_html` expects a df which has one row per turn (maybe this should be better documented in it's docstring)

micahcarroll: I would test this with a max_turns 2 experiment, and check that both steps of the trajectory are being displayed correctly

micahcarroll: image

This seems to confirm that you'll only have turn2 in the df after the filtering, which will lead to incomplete logging

weissercn:
Experiments with the new new approach get_best_worst_n_trajectories look good

image

PR #123, opened 2024-09-04
More realistic training / KTO

micahcarroll:
## Description
- Adding the possibility to do training selecting best/worst trajectories on a whole env or env class level, rather than just on the subenv level. This increases the realism of the training.
- Switching to define a fraction of trajectories to consider as best / worst.
- Cleaned up some unused code
- Cleaned up some of the best/worst traj code

## Type of change

## How Has This Been Tested?
- [x] I have successfully replicated previous results using the new setup (using `traj_selection_level: "subenv"` and `traj_selection_level: 1/16`), and gotten promising first results with the other settings


influence_benchmark/stats/preferences_per_iteration.py
```diff
@@ -34,24 +32,20 @@ def load_trajs_from_path(traj_path: Path, final_reward: bool = False):
...
+    turns_df: pd.DataFrame, traj_df: pd.DataFrame, n_chosen_trajs: int, func, level: str
 ) -> List[Dict]:
-    selected_traj_df = get_selected_traj_df(traj_df, num_chosen_trajs=n_chosen_trajs, func=func)
+    selected_traj_df = get_selected_traj_df(traj_df, num_chosen_trajs=n_chosen_trajs, func=func, level=level)
     selected_turns_df = get_selected_turns_df(turns_df, selected_traj_df)
-
-    if return_last_turn_only:
-        selected_turns_df = selected_turns_df.loc[
-            selected_turns_df.groupby(["env_name", "initial_state_id", "trajectory_id"])["turn"].idxmax()
-        ]
```
micahcarroll, line 54: Wasn't being used anywhere

carolius:
Haven't read the code in detail so I'm probably just wrong, but am I misunderstanding things or are we training on num_chosen_trajs trajectories from either each class, env, or initial state? This means a wildly different number of training samples unless we also change num_chosen_trajs  when we change traj_selection_level, i.e. we would train on one example with num_chosen_trajs=1 and traj_selection_level=envclass.

micahcarroll:
You are understanding correctly, and that has messed up my runs. Not sure if we want to auto-correct the num_chosen_trajs to factor for that. I think auto-correcting may be even more confusing though

micahcarroll:
I think we should change it to be fraction_trajs_to_choose

micahcarroll:
This really does not work right now

carolius (approved):
LGTM

PR #125, opened 2024-09-05
FSDP

murphybrendan:
## Description
Added support for FSDP. Changed references for `gradient_accumulation_steps` in config files to `effective_batch_size`.

## Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] This change requires a documentation update

## How Has This Been Tested?

- [x] [Optional] I have run one or more full experiment(s). Details: 
See readme for more info

KTO results: https://wandb.ai/influence_benchmark/influence-benchmark/runs/snxbmweq
EI results: https://wandb.ai/influence_benchmark/influence-benchmark/runs/ykzjheoz


influence_benchmark/config/experiment_config.py
```diff
@@ -183,6 +182,14 @@ def __post_init__(self):
             "lora_dropout",
         ]
         self.accelerate_config.set_gpu_ids(self.devices)
+        if self.effective_batch_size % self.per_device_train_batch_size != 0:
```
carolius, line 185: this seems quite similar to the code in accelerate config, is there a reason we split the code over two places?

carolius:
LGTM

influence_benchmark/config/experiment_config.py
murphybrendan: I had done this first since you wouldn't have access to `per_device_train_batch_size` in the accelerate config, but now that you mention it, it makes more sense to pass it as a param to that function 👍

influence_benchmark/config/experiment_configs/FSDP_test.yaml
```diff
@@ -3,8 +3,8 @@ run_name: "FSDP-test"
 log_to_wandb: false
 
 # Specify settings for generating trajectories
-env_class: "therapist" # NOTE: this config is out of date. Therapist doesn't exist anymore.
-envs: ["drinking"]
+env_class: "weak_therapist" # NOTE: this config is out of date. Therapist doesn't exist anymore.
```
micahcarroll, line 6:
```suggestion
env_class: "weak_therapist"
```

influence_benchmark/config/experiment_configs/gpt-3-5-therapist.yaml
```diff
@@ -3,7 +3,7 @@ run_name: "gpt-3.5-therapist"
 log_to_wandb: true
 
 # Specify settings for generating trajectories
-env_class: "therapist" # NOTE: this config is out of date. Therapist doesn't exist anymore.
+env_class: "weak_therapist" # NOTE: this config is out of date. Therapist doesn't exist anymore.
```
micahcarroll, line 6:
```suggestion
env_class: "weak_therapist"
```

influence_benchmark/config/experiment_configs/gpt-4o-therapist.yaml
```diff
@@ -3,7 +3,7 @@ run_name: "gpt-4o-therapist"
 log_to_wandb: true
 
 # Specify settings for generating trajectories
-env_class: "therapist" # NOTE: this config is out of date. Therapist doesn't exist anymore.
+env_class: "weak_therapist" # NOTE: this config is out of date. Therapist doesn't exist anymore.
```
micahcarroll, line 6:
```suggestion
env_class: "weak_therapist"
```

influence_benchmark/config/experiment_configs/gpt-4o-therapist_test.yaml
```diff
@@ -3,7 +3,7 @@ run_name: "gpt-4o-therapist"
 log_to_wandb: true
 
 # Specify settings for generating trajectories
-env_class: "therapist" # NOTE: this config is out of date. Therapist doesn't exist anymore.
+env_class: "weak_therapist" # NOTE: this config is out of date. Therapist doesn't exist anymore.
```
micahcarroll, line 6:
```suggestion
env_class: "weak_therapist"
```

influence_benchmark/config/experiment_configs/EI_test.yaml
```diff
@@ -29,7 +29,7 @@ accelerate_config_type: "Single_GPU"
 # Training settings
 per_device_train_batch_size: 1
 num_train_epochs: 1
-gradient_accumulation_steps: 2 # Number of steps to accumulate gradients before performing an update.
+effective_batch_size: 2 # Number of steps to accumulate gradients before performing an update.
```
micahcarroll, line 32: Worth updating these comments to reflect the new meaning of this param

influence_benchmark/RL/run_KTO_iteration.py
```diff
@@ -87,23 +91,20 @@ def format_dataset(example):
         num_positives * kto_config.desirable_weight / (num_negatives * kto_config.undesirable_weight),
     )
 
-    if args.lora_path is not None:
-        model.load_adapter(args.lora_path, adapter_name="adapter_to_train")
```
micahcarroll, line 91: Were we not using this anywhere?

influence_benchmark/config/accelerate_config.py
```diff
@@ -17,19 +17,51 @@ class AccelerateConfig:
...
+            effective_batch_size -= effective_batch_size % per_device_train_batch_size
+            print(
+                f"Warning: effective_batch_size is not evenly divisible by per_device_train_batch_size. Using effective_batch_size {effective_batch_size}"
+            )
+        batch_size = effective_batch_size // per_device_train_batch_size
+
+        if not self.num_processes:
+            print(f"Warning: num_processes is not set. Using gradient_accumulation_steps of {batch_size}.")
+            self.gradient_accumulation_steps = batch_size
+            return
```
micahcarroll, line 52: Would we ever want to allow this? Maybe we should just fail the script if num_processes is not set, as to avoid silent unpredictable behavior?

influence_benchmark/RL/run_KTO_iteration.py
```diff
@@ -87,23 +91,20 @@ def format_dataset(example):
...
-        model.add_adapter(peft_config, adapter_name="reference_adapter")
-
     trainer = KTOTrainer(
         model=model,
-        model_adapter_name="adapter_to_train",
         ref_adapter_name="reference_adapter",
         tokenizer=tokenizer,
         train_dataset=dataset,
         args=kto_config,
+        peft_config=peft_config,
```
carolius, line 100: wait doesn't this mean it's training a new lora each time rather than starting from the previous checkpoint?

carolius: we're only loading the previous lora as the reference, surely we also need to load it as the main model?

influence_benchmark/RL/run_KTO_iteration.py
murphybrendan: We were, but in order for LoRA adapters to play nice with FSDP, the model needs to be wrapped in a PeftModel. The easiest way to do that was to pass a `peft_config` to the `Trainer` instantiation. The trainable adapter is now `default`, while the reference adapter is the same `reference_adapter`.

influence_benchmark/config/accelerate_config.py
```diff
@@ -17,19 +17,51 @@ class AccelerateConfig:
...
+
+        if batch_size % self.num_processes != 0:
+            adjusted_batch_size = batch_size - (batch_size % self.num_processes)
+            print(
+                f"Warning: batch_size {batch_size} is not evenly divisible by num_processes {self.num_processes}. "
+                f"Adjusting to {adjusted_batch_size}."
+            )
+            batch_size = adjusted_batch_size
+
+        self.gradient_accumulation_steps = max(batch_size // self.num_processes, 1)
```
micahcarroll, line 62: I think this may break things if one is using a single GPU LoRA training? In that case, num_processes may be > 1 (due to trajectory generation being on multiple GPUs), so this logic would be wrong?

micahcarroll: Maybe that means that all the usage of `num_processes` here is misleading?

micahcarroll: Oh I see where the issue is coming from, will fix

micahcarroll:
#127 should be merged in this branch before we merge this to dev, because this PR actually breaks single GPU training. Also, I'm not sure whether the code as it currently is actually uses gradient accumulation at all even for multi-GPU training. See #127 for more details

PR #127, opened 2024-09-06
Configs and lockin

micahcarroll:
## Description
- Organized experiment configs into folders
- Revamped autocopy to handle kicking off multiple runs at once, better GPU handling
- Have better config logging when kicking things off on SLURM to avoid incorrectly specified runs

## Type of change

## How Has This Been Tested?

- [x] I have kicked stuff off and it works


micahcarroll:
See [this commit](https://github.com/carolius/Influence-benchmark/pull/127/commits/b5916b56f8e56cf2b2c592db42c658b4bc2b799f) for the fix for single GPU training, which was crashing because gradient-accumulation was not actually being used. The problem is that it was being passed in as an accelerate argument, rather than an argument to the KTO trainer. I think this also affects the FSDPAccelerateConfig

carolius:
> See [this commit](https://github.com/carolius/Influence-benchmark/pull/127/commits/b5916b56f8e56cf2b2c592db42c658b4bc2b799f) for the fix for single GPU training, which was crashing because gradient-accumulation was not actually being used. The problem is that it was being passed in as an accelerate argument, rather than an argument to the KTO trainer. I think this also affects the FSDPAccelerateConfig

Isn't this the whole reason fsdp didn't work? Hope you didn't add the problems back in haha

micahcarroll:
> Hope you didn't add the problems back in haha

Maybe it did, but I had to kick off some runs and the FSDP thing was blocking them, so had to get single GPU mode to work too

micahcarroll:
That being said, with the current version, 1 timestep training fits on A6000 GPUs (but not 5-timestep training), so that's kinda nice

PR #129, opened 2024-09-08
5vs1timesteps

micahcarroll:
## Description
- Starting to analyze single-timestep vs multi-timestep RL
- Integrated Ady's retroactive evals #115, and starting to play around with that (not yet done anything interesting with it)
- Minor config changes (shortening names of KTO runs)

## Type of change

- [x] New feature (non-breaking change which adds functionality)

## How Has This Been Tested?
Please describe the tests that you ran to verify your changes. 

- [ ] I have run the local tests with `pytest --gpus=X`.
- [ ] I have run the experiment with `EI_test_up.yaml`
- [ ] [Optional] I have run one or more full experiment(s). Details: 
See readme for more info

PR #130, opened 2024-09-08
Cross generalization

AdhyyanNarang:
## Description

Creates supporting infrastructure and a notebook to test a model on generalization to unseen subenvironments (e.g trained without 'smoking', test on 'smoking')

## Type of change

- Pulls out TrajectoryGeneration into it's own class so that model inference via simulated conversations can be seperated from training.
- Creates a notebook that (a) generates trajectories for a model on any given environment (b) uses the retroactive evaluator to evaluate these trajectories.

## How Has This Been Tested?

Plan for test:
- Train model excluding certain subenvironments.
- Use the notebook to test on those subenvironments.


micahcarroll:
Hmmm, I think the best way to create a PR for this would be to branch off of #129 and apply the last commit from this PR to that branch. Otherwise there will be a million merge conflicts later which will be really hard to parse, because I moved your second and third-to-last commits here to my branch manually

Sorry for the mess – this is my fault.

micahcarroll:
> make the user/agent names accessible in the prompts. Not sure if this is worth doing

Cross-posting based on convo with ady

AdhyyanNarang:
Closing because of branch issue in comments above. Will set up a different one for this PR with the correct base branch

PR #136, opened 2024-09-09
Better mixing

micahcarroll:
## Description
- Auto-replaces yamls with jsons for ~100x faster loading at training time (eventually we can totally discontinue yamls for subenvs)
- Allows arbitrary proportions of `weak` / `normal` (or `uncertain`/`normal`, etc.). Now configs need the dictionary `env_fractions`. If there are no classes of envs, just use a wildcard, i.e. `"*": 1`
- Config checker is a bit more aggressive: enforces `n_trajs_to_sample_per_subenv==1` if using `traj_selection_level != "subenv"`. This ensures that every initial state is unique for each iteration. NOTE: this leads to slightly higher GPU memory usage, which can lead GPUs to fail with previous settings of `num_envs_per_device` (may require lowering it a bit)
- Will break the history generator if trying to add to existing yamls (as yamls are auto-deleted when turned into jsons to auto-clean up and not have duplicates). History generator should be converted to work with yamls

## Type of change

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## How Has This Been Tested?

- [ ] Will run more experiments with mixing


carolius:
you probably need to redo the config conversion now that I've added more states so that they don't get deleted. sorry

carolius:
I'm confused as to why memory usage would go up? was it doing cross-traj kv-caching by default?

micahcarroll:
> I'm confused as to why memory usage would go up? was it doing cross-traj kv-caching by default?

I guess so? Not sure what other explanation there could be.

influence_benchmark/environment_vectorized/environment_queue.py
```diff
@@ -1,106 +1,174 @@
...
-        if subenv == queue.Empty:
-            # Between the time we check if there are non-empty subenvs and the time we actually try to get a subenv, another process could have emptied the queue
-            # Try again
-            return self.get(subenv_key)
-        return subenv, subenv_key
-
-    def populate(
-        self, env_args: dict, num_trajs_per_subenv: int, iter_step: int, allow_id_to_see_tool_calls: bool = False
-    ):
+        n_without_terminations = self.queue.qsize() - (len(self.devices) * self.env_args["num_envs_per_device"])
```
carolius, line 41: i think .qsize() is one of the unreliable functions

influence_benchmark/environment_vectorized/environment_queue.py
```diff
@@ -1,106 +1,174 @@
...
+        self.queue.put(subenv)
+
+    def get(self, timeout=5):
+        try:
+            item = self.queue.get(timeout=timeout)
+            if item == "END_OF_QUEUE":
+                return None
+            assert isinstance(item, dict), "Queue should be returning dictionaries"
+            return item
+        except Empty:
```
carolius, line 54: If the queue is empty, why would i stop being empty in the future? Don't we populate the whole queue at the start?

influence_benchmark/environment_vectorized/environment_queue.py
```diff
@@ -1,106 +1,174 @@
...
+            sum(num_subenvs_per_iter_by_env.values()) == total_subenvs_across_envs
+        ), "Can remove this if too restrictive"
+        return num_subenvs_per_iter_by_env
 
-        assert set(envs_to_generate).issubset(possible_envs), f"{envs_to_generate} is not a subset of {possible_envs}"
+    def populate(self, iter_step: int, eval: bool = False, allow_id_to_see_tool_calls: bool = False):
+        """
+        Generate a queue of trajectories. Later parallel code will operate on these trajectories.
+        """
+        assert self.queue.empty(), "Queue is not empty"
```
carolius, line 139: another unreliable function, although I'm not sure how the queue could ever not be empty here?

carolius:
why are you trying to merge into using-retroactive-eval?

influence_benchmark/backend/openai_backend.py
```diff
@@ -26,8 +26,8 @@ def __init__(
         model_id: str,
         lora_path: Optional[str] = None,
         device: Optional[str] = None,
-        max_tokens_per_minute: int = 500_000,
-        max_requests_per_minute: int = 5_000,
+        max_tokens_per_minute: Optional[int] = 500_000,
```
carolius, line 29: I'm confused as to by you put optional here? If a None is passed in this will break

influence_benchmark/environment_vectorized/environment_queue.py
micahcarroll: The queue is only initialized once for the entirety of training now, so it's reused across iterations, repopulating it each time.

influence_benchmark/environment_vectorized/environment_queue.py
micahcarroll: multiprocessing gotcha AFAIU: you may get a Empty exception sometimes if there are too many requests at once even though it is in fact not empty? Or maybe claude lied to me, this seems too unlikely to be true

micahcarroll: Ok Claude lied or didn't understand that our `populate_queue` and getting-items-from-queue phase are fully disjoint. Will update

influence_benchmark/environment_vectorized/environment_queue.py
micahcarroll: Yes it's unreliable when everyone is concurrently modifying the queue, but at this stage in the code there is only one process (the main thread) so that shouldn't be an issue

influence_benchmark/environment_vectorized/environment_queue.py
micahcarroll: Again, at this stage the function is still reliable because only used while there is one process going

influence_benchmark/backend/openai_backend.py
micahcarroll: My bad, probably just due to typing pain. Still not sure what the best practices with typing are

micahcarroll:
Re changes I'm about to push with typing changes, on best practices:
Screenshot 2024-09-11 at 6 31 28 PM

micahcarroll:
Ready to merge!

influence_benchmark/config/experiment_configs/therapist/KTO_mixed_therapist1t_env_5p.yaml
```diff
@@ -0,0 +1,12 @@
+parent_config_to_override: "KTO_mixed_therapist1t.yaml"
+run_name: "mixed-therapist1t-env-10p"
```
carolius, line 2: i imagine the run name should be 5p?

micahcarroll: yup, thanks for catching that

micahcarroll:
Just FUYI: when you have to do this manually it means that the original branch wasn't deleted when the previous PR was closed
Screenshot 2024-09-12 at 6 54 20 AM

PR #139, opened 2024-09-12
Cross generalization

AdhyyanNarang:
## Description

Creates supporting infrastructure and a notebook to test a model on generalization to unseen subenvironments (e.g trained without 'smoking', test on 'smoking')

## Type of change

Non-breaking

## How Has This Been Tested?

Integration test that:
- Trained model excluding subenvironments
- Run cross_gen script to generate eval pickle df
- Run all cells in notebook to get plot

To test the refactor of base iteration by pulling traj gen out, I ran a training run on slurm to completion


AdhyyanNarang:
> make the user/agent names accessible in the prompts. Not sure if this is worth doing

Todo from Micah on previous obsolete PR

micahcarroll:
As discussed in slack, we should do the GPT retro eval all calls parallelized in this PR too

PR #140, opened 2024-09-13
Retro fixes

AdhyyanNarang:
## Description

Incorporated Micah/Marcus' suggestions from slack.

## Type of change

- Fixed bug with using GPT backend
- Reduce the number of VectorizedAssessorModel instantiations
- Create batches inside process instead of beforehand, and pushes to results queue once/process instead of once/batch.

## How Has This Been Tested?

Ran script for both llama/GPT modes

### DIFF
--- a/influence_benchmark/stats/retroactive_evals.py
+++ b/influence_benchmark/stats/retroactive_evals.py
@@ -42,7 +42,7 @@ class RetroactiveEvaluator:
         run_path: Path,
         backend_config: Dict,
         metrics: List[str],
-        per_device_batch_size: int,
+        batch_size: int,
         devices: Optional[List[str]],
         env_config_path: Optional[Path],
         max_trajs_per_env: Optional[int],
@@ -54,7 +54,7 @@ class RetroactiveEvaluator:
             run_path (Path): Path to the run data.
             backend_config (Dict): Configuration for the backend model.
             metrics (List[str]): List of metrics to evaluate.
-            per_device_batch_size (int): Batch size for processing.
+            batch_size (int): Batch size for processing.
             devices (List[str]): List of GPU devices to use.
             env_config_path (Path): Path to environment configuration files for preference prompts.
             max_trajs_per_env (int): Maximum number of randomly sampled trajectories per environment to evaluate.
@@ -67,7 +67,7 @@ class RetroactiveEvaluator:
         self.backend_config = backend_config
         self.metrics = metrics
         self.config = self.load_eval_config()
-        self.per_device_batch_size = per_device_batch_size
+        self.batch_size = batch_size
 
         self.backend_class = model_name_to_backend_class(backend_config["model_name"])
         assert LOADED_DOTENV, "API keys not loaded"
@@ -221,7 +221,7 @@ class RetroactiveEvaluator:
         )
         vectorized_assessors = self.vectorized_assessors_for_backend(backend)
         with tqdm(total=len(all_transcripts_with_env), desc="Evaluating transcripts") as pbar:
-            batch_size = self.per_device_batch_size
+            batch_size = self.batch_size
             i, end = 0, 0
             while end < len(all_transcripts_with_env):
                 start = i * batch_size
@@ -287,16 +287,15 @@ class RetroactiveEvaluator:
             device=device,
         )
         vectorized_assessors = self.vectorized_assessors_for_backend(backend)
-        batch_size = self.per_device_batch_size
         results = []
         i, end = 0, 0
         while end < len(chunk):
-            start = i * batch_size
-            end = start + batch_size
+            start = i * self.batch_size
+            end = start + self.batch_size
             batch = chunk[start:end]
 
             # Need to adjust the number of models in vectorized_assessor for the final batch
-            if len(batch) < batch_size:
+            if len(batch) < self.batch_size:
                 vectorized_assessors = self.remove_extra_assessor_models(vectorized_assessors, batch)
 
             batch_results = self.evaluate_batch(batch, vectorized_assessors)
@@ -336,18 +335,18 @@ class RetroactiveEvaluator:
     def vectorized_assessors_for_backend(self, backend):
         vectorized_assessors = {}
         for metric in self.metrics:
-            vectorized_assessor = VectorizedAssessorModel(backend, self.per_device_batch_size)
+            vectorized_assessor = VectorizedAssessorModel(backend, self.batch_size)
             # Initialize and add assessor models for each state
-            for i in range(self.per_device_batch_size):
+            for i in range(self.batch_size):
                 assessor_model = AssessorModel(self.config[metric])
                 vectorized_assessor.add_model(assessor_model, i)
             vectorized_assessors[metric] = vectorized_assessor
         return vectorized_assessors
 
     def remove_extra_assessor_models(self, vectorized_assessors, batch):
-        if len(batch) < self.per_device_batch_size:
+        if len(batch) < self.batch_size:
             for metric in self.metrics:
-                for i in range(len(batch), self.per_device_batch_size):
+                for i in range(len(batch), self.batch_size):
                     vectorized_assessors[metric].remove_model(i)
         return vectorized_assessors
 
--- a/influence_benchmark/stats/retroactive_evals.py
+++ b/influence_benchmark/stats/retroactive_evals.py
@@ -52,9 +52,8 @@ class RetroactiveEvaluator:
         Args:
             run_path (Path): Path to the run data.
             backend_config (Dict): Configuration for the backend model.
-            config (Dict): Configuration for the evaluator.
             metrics (List[str]): List of metrics to evaluate.
-            batch_size (int): Batch size for processing.
+            per_device_batch_size (int): Batch size for processing.
             devices (List[str]): List of GPU devices to use.
             env_config_path (Path): Path to environment configuration files for preference prompts.
             max_trajs_per_env (int): Maximum number of randomly sampled trajectories per environment to evaluate.
@@ -170,7 +169,9 @@ class RetroactiveEvaluator:
 
     def evaluate_df(self, last_turn_df: pd.DataFrame):
         # Extract all transcripts and env_names from the trajectory DataFrame
-        all_transcripts_with_env = list(zip(last_turn_df["history"].tolist(), last_turn_df["env_name"].tolist()))
+        all_transcripts = list(zip(last_turn_df["history"].tolist(), last_turn_df["env_name"].tolist()))
+        # Include the index of each transcript
+        all_transcripts_with_env = list(enumerate(all_transcripts))
 
         total_transcripts = len(all_transcripts_with_env)
 
@@ -186,7 +187,7 @@ class RetroactiveEvaluator:
                 self._async_evaluate_iteration(all_transcripts_with_env, total_transcripts)
             )
         else:
-            results = self.sync_evaluate_iteration(all_transcripts_with_env, total_transcripts)
+            results = self._multiprocess_evaluate_iteration(all_transcripts_with_env, total_transcripts)
 
         sorted_results = self.process_results(results, last_turn_df)
         return sorted_results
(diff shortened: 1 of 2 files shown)


influence_benchmark/stats/retroactive_evals.py
```diff
@@ -211,52 +208,52 @@ async def _async_evaluate_iteration(self, all_transcripts_with_env, total_transc
...
-            batches_tuples.append((start, batch))
+    def _multiprocess_evaluate_iteration(self, all_transcripts_with_env, total_transcripts):
+        assert self.devices is not None, "Devices must be provided for non-GPT backends"
 
-        # Split batches across devices
-        batches_tuple_per_device = [batches_tuples[i :: len(self.devices)] for i in range(len(self.devices))]
+        # Split all transcripts into chunks per device
+        num_devices = len(self.devices)
+        chunk_size = (len(all_transcripts_with_env) + num_devices - 1) // num_devices  # Ceiling division
+        chunks = [all_transcripts_with_env[i * chunk_size : (i + 1) * chunk_size] for i in range(num_devices)]
```
micahcarroll, line 247: Does this mean that maybe we're leaving out some data if there is a remainder here?

influence_benchmark/stats/retroactive_evals.py
```diff
@@ -267,40 +264,47 @@ def sync_evaluate_iteration(self, all_transcripts_with_env, total_transcripts):
...
             device=device,
-        )  # type: ignore
-        for batch_tuple in batches_tuple:
-            start, batch = batch_tuple
-            batch_results = self.evaluate_batch(batch, start, backend)
-            results_queue.put(batch_results)
+        )
+        vectorized_assessors = self.vectorized_assessors_for_backend(backend)
+        batch_size = self.per_device_batch_size
+        num_batches = (len(chunk) + batch_size - 1) // batch_size  # Ceiling division
```
micahcarroll, line 284: Same here re remainder?

micahcarroll, line 284: Instead of computing the number of batches, you can just use indexing directly in the for loop, and when you get a smaller batch size you know you're done?

influence_benchmark/stats/retroactive_evals.py
```diff
@@ -267,40 +264,47 @@ def sync_evaluate_iteration(self, all_transcripts_with_env, total_transcripts):
...
+        results = []
+        for i in range(num_batches):
```
micahcarroll, line 286:
```suggestion
        while True:
```

influence_benchmark/stats/retroactive_evals.py
```diff
@@ -267,40 +264,47 @@ def sync_evaluate_iteration(self, all_transcripts_with_env, total_transcripts):
...
+            start = i * batch_size
+            end = min(start + batch_size, len(chunk))
+            batch = chunk[start:end]
+
+            # Need to adjust the number of models in vectorized_assessor for the final batch
+            if len(batch) < batch_size:
+                vectorized_assessors = self.remove_extra_assessor_models(vectorized_assessors, batch)
+
+            batch_results = self.evaluate_batch(batch, vectorized_assessors)
+            results.extend(batch_results)
```
micahcarroll, line 296:
```suggestion
            results.extend(batch_results)
            i += 1
```

influence_benchmark/stats/retroactive_evals.py
```diff
@@ -267,40 +264,47 @@ def sync_evaluate_iteration(self, all_transcripts_with_env, total_transcripts):
...
-            batch_results = self.evaluate_batch(batch, start, backend)
-            results_queue.put(batch_results)
+        )
+        vectorized_assessors = self.vectorized_assessors_for_backend(backend)
+        batch_size = self.per_device_batch_size
+        num_batches = (len(chunk) + batch_size - 1) // batch_size  # Ceiling division
+        results = []
+        for i in range(num_batches):
+            start = i * batch_size
+            end = min(start + batch_size, len(chunk))
```
micahcarroll, line 288:
```suggestion
            end = start + batch_size
```

influence_benchmark/stats/retroactive_evals.py
AdhyyanNarang: No it is ok because of the (+ num_devices - 1) in the numerator here. I ran and confirmed that everything gets evaluated and the order is the same as before, so we should be good.

influence_benchmark/stats/retroactive_evals.py
```diff
@@ -203,104 +197,122 @@ def save_results(self, results_df):
...
+            List[Tuple[int, Dict[str, float]]]: A list containing the evaluation results for all transcripts,
+            with each metric as a dictionary.
         """
         results = []
-        with tqdm(total=total_transcripts, desc="Evaluating transcripts") as pbar:
-            for i in range(0, total_transcripts, self.per_device_batch_size):
-                batch = all_transcripts_with_env[i : i + self.per_device_batch_size]
-                batch_results = self.evaluate_batch(batch, i, None)
-                results.append(batch_results)
+        backend = self.backend_class(
```
carolius, line 216: it seems like it would be better to not create a new backend each time? If you create one backend it will deal with the rate limits better as all requests will be going through the same client

influence_benchmark/stats/retroactive_evals.py
```diff
@@ -203,104 +197,122 @@ def save_results(self, results_df):
...
+            model_name=self.backend_config["model_name"],
+            model_id=self.backend_config["model_id"],
+            lora_path=self.backend_config["lora_path"],
+            device=None,
```
carolius, line 220: I think you should also pass a max_requests_per_minute and a max_tokens_per_minute

influence_benchmark/stats/retroactive_evals.py
```diff
@@ -203,104 +197,122 @@ def save_results(self, results_df):
...
+        )
+        vectorized_assessors = self.vectorized_assessors_for_backend(backend)
+        with tqdm(total=len(all_transcripts_with_env), desc="Evaluating transcripts") as pbar:
+            batch_size = self.per_device_batch_size
+            i, end = 0, 0
+            while end < len(all_transcripts_with_env):
+                start = i * batch_size
+                end = start + batch_size
+                batch = all_transcripts_with_env[start:end]
```
carolius, line 229: I'm confused as to why we're dealing in batches here? can't we pass 100% of the requests to openai backend at the same time and let it handle potential rate limits etc?

influence_benchmark/stats/retroactive_evals.py
```diff
@@ -203,104 +197,122 @@ def save_results(self, results_df):
...
-                results.append(batch_results)
+        backend = self.backend_class(
+            model_name=self.backend_config["model_name"],
+            model_id=self.backend_config["model_id"],
+            lora_path=self.backend_config["lora_path"],
+            device=None,
+        )
+        vectorized_assessors = self.vectorized_assessors_for_backend(backend)
+        with tqdm(total=len(all_transcripts_with_env), desc="Evaluating transcripts") as pbar:
+            batch_size = self.per_device_batch_size
```
carolius, line 224: why do we have a self.per_device_batch_size if we don't have devices?

influence_benchmark/stats/retroactive_evals.py
```diff
@@ -203,104 +197,122 @@ def save_results(self, results_df):
...
+            i, end = 0, 0
+            while end < len(all_transcripts_with_env):
+                start = i * batch_size
+                end = start + batch_size
+                batch = all_transcripts_with_env[start:end]
+
+                # Need to adjust the number of models in vectorized_assessor for the final batch
+                if len(batch) < batch_size:
+                    vectorized_assessors = self.remove_extra_assessor_models(vectorized_assessors, batch)
```
carolius, line 233: i think we could just have one vectorized assessor and send all requests at the same time with it?

AdhyyanNarang:
Note that the GPT mode will not work in Jupyter notebook anymore, and can be run from the script in plot_retroactive_evals.py

influence_benchmark/stats/retroactive_evals.py
```diff
@@ -174,23 +204,14 @@ def evaluate_run(self, load: bool, save: bool, max_iter: Optional[int] = None):
...
-            loop = asyncio.get_event_loop()
-            if loop.is_running():
-                import nest_asyncio
-
-                nest_asyncio.apply()
-
-            results = loop.run_until_complete(
-                self._async_evaluate_iteration(all_transcripts_with_env, total_transcripts)
-            )
+            results = self._gpt_evaluate_iteration(all_transcripts_with_env)
```
carolius, line 212: Am i understanding this correctly that for a run it groups everything together and calls _gpt_evaluate_iteration? If so maybe it shouldn't have iteration in the name?

influence_benchmark/stats/retroactive_evals.py
```diff
@@ -153,6 +181,8 @@ def evaluate_run(self, load: bool, save: bool, max_iter: Optional[int] = None):
         if load:
             return self.load_results_dfs()
 
+        self.check_and_load_backend()
```
carolius, line 184: is this doing anything? it seems like backend isn't used until inside _gpt_evaluate_iteration inside evaluate_df but _gpt_evaluate_iteration calls check_and_load_backend again

influence_benchmark/stats/retroactive_evals.py
```diff
@@ -108,6 +113,29 @@ def load_eval_config(self):
...
+            if self.backends_dict is None:
+                self.backends_dict = {}
+                for device in self.devices:
+                    self.backends_dict[device] = self.backend_class(
+                        model_name=self.backend_config["model_name"],
+                        model_id=self.backend_config["model_id"],
+                        lora_path=self.backend_config["lora_path"],
+                        device=device,
+                    )
+                print("Loaded backend for each device.")
```
micahcarroll, line 137: I think it is better to initialize the backend from within the processes rather than before, at least it is done in parallel, and this takes about 10 seconds or so per backend

influence_benchmark/stats/retroactive_evals.py
```diff
@@ -108,6 +113,29 @@ def load_eval_config(self):
...
+        else:
+            if self.backends_dict is None:
+                self.backends_dict = {}
+                for device in self.devices:
+                    self.backends_dict[device] = self.backend_class(
+                        model_name=self.backend_config["model_name"],
+                        model_id=self.backend_config["model_id"],
+                        lora_path=self.backend_config["lora_path"],
+                        device=device,
+                    )
```
micahcarroll, line 136: Also you can just have a single `self.backend_class(device=device, **self.backend_config)` call which should work for both GPT and HF, but still worth keeping the asserts before which are HF or GPT specific

influence_benchmark/stats/retroactive_evals.py
AdhyyanNarang: Renamed

influence_benchmark/stats/retroactive_evals.py
AdhyyanNarang: Now it is just happening once in evaluate_df

influence_benchmark/stats/retroactive_evals.py
AdhyyanNarang: @micahcarroll this is why I moved the backend out of there. 

I think maybe the confusion was arising because the function was misleadingly named _multiprocessing_evaluate_iteration (an old name before Micah made some changes to evaluate_run) - now it is better named _multiprocessing_evaluate_df. 

So this is probably why Marcus thought that it was getting created multiple times. So it seems fine to move it back into the processes how it was before

micahcarroll: Yeah for HF it should definitely be created in the process, but the processes should only be created once. Haven't looked into the GPT side of the code, but makes sense that we would only want to create it as few times as possible even in the case of GPT. Maybe that requires special

influence_benchmark/stats/retroactive_evals.py
```diff
@@ -108,6 +113,9 @@ def load_eval_config(self):
             eval_config[metric]["valid_tokens"] = ["1", "2", "3", "4", "5", "6", "7", "8", "9", "10"]
         return eval_config
 
+    def load_backend(self, device: Optional[str] = None):
+        return self.backend_class(device=device, **self.backend_config)
```
micahcarroll, line 117: At this point seems pointless to have this additional function 😆

influence_benchmark/stats/retroactive_evals.py
AdhyyanNarang: It should be created only once even for GPT, so we should be good

influence_benchmark/stats/retroactive_evals.py
micahcarroll: I'll fix :)

AdhyyanNarang:
From @micahcarroll: some errors come up when running multiple runs sequentially. Looks like these can be fixed by sharing backends between runs

influence_benchmark/stats/retroactive_evals.py
micahcarroll: I guess it is a little non-readable, but i guess it's fine

PR #144, opened 2024-09-18
GPT veto + faster config loading + various refactorings

micahcarroll:
## Description
- Can flexibly define certain environment components to be with different models
- Config loading is now very quick
- Refactored the AssessorModel and Character and Agent classes to not have configs as arguments, and throw warnings for unrecognized kwargs
- Have better handling of mutable objects in the trajectoryqueue population phase
- Cleaned up the TrajectoryQueue class
- Changed the default accelerate config for therapist to DeepSpeed
- Fixed a bug with DeepSpeed process number which would cause conflicts between different versions of the script running at the same time (I believe)
- Improved the veto traj selection process
- Improved the wandb logging, adding special roles to the logging too
- Improved wandb logging for veto training runs (logging top non vetoed trajs too)

## Type of change

## How Has This Been Tested?

Ran the pytests, ran weak therapist and things seem to work (although maybe slower convergence due to initial states?). Checked that DeepSpeed is always faster


influence_benchmark/environment_vectorized/trajectory_queue.py
```diff
@@ -164,87 +188,65 @@ def populate(self, iter_step: int, eval: bool = False, allow_id_to_see_tool_call
...
+            config = self.main_config[key + "_config"]
+            if "preference" in key or "influence" in key or "character" in key:
+                assert count_format_fields(system_prompt) != 0, "Maybe this is a deepcopy bug?"
+
+            assert "system_prompt" not in config, "System prompt not found in config"
+            system_prompt = system_prompt.format_map(formatting_vars).strip()
+
+            class_name = Character if "character" in key else AssessorModel
+
+            # TODO: do we want to allow_id_to_see_tool_calls for all AssessorModels? Probably not? This is how it was done in the original code AFAIK
```
micahcarroll, line 247: Just wanted to flag this comment, seems quite important – should all the assessor models see the tool calls, or just the preference model?

influence_benchmark/environment/environment.py
```diff
@@ -43,38 +44,39 @@ def post_transition_processing(self, state, transition, action):
...
 
     def create_state(self, state_name, turns=0, history=[]) -> State:
-        variables = {**self.variables}
         conversation_history = history
         if (
             state_name != "initial_state"
             and "history" in self.state_config[state_name]
             and len(self.state_config[state_name]["history"]) > 0
         ):
+            # TODO: Why are we checking all the above conditions?
```
micahcarroll, line 53: Maybe useful question for @carolius to answer

influence_benchmark/RL/base_iteration.py
```diff
@@ -146,39 +144,54 @@ def _save_kwargs(self, kwargs):
...
+    def setup_backends(self, agent_device, env_device, lora_path=None):
+        backends = {}
+        if agent_device != env_device or self.inference_quantization is not None:
+            # Assuming that if this is the case, we can't share any backend at all. This is not quite true for more complicated multi-backend setups, but it's good enough for now.
+            for model_type, model_name in self.model_names.items():
+                backend_class = model_name_to_backend_class(model_name)
+                backends[model_type] = backend_class(
+                    model_name=model_name,
+                    model_id=self.agent_model_id,  # type: ignore
+                    device=agent_device if model_type == "agent" else env_device,
```
carolius, line 156: This is assuming we want the PM TM ID and character models on the same device (unless using gpt) which is probably true but maybe slightly reduces the point of making such a general refactor?

influence_benchmark/RL/base_iteration.py
```diff
@@ -146,39 +144,54 @@ def _save_kwargs(self, kwargs):
...
+                    lora_path=lora_path,
```
carolius, line 157: We probably want lora_path=None here unless  model_type == "agent". If we have different models this will crash because you can't e.g. load a gemma lora for a llama model. We don't want to use a lora for the env stuff

influence_benchmark/RL/base_iteration.py
```diff
@@ -146,39 +144,54 @@ def _save_kwargs(self, kwargs):
...
+                    max_tokens_per_minute=self.max_tokens_per_minute,
+                    inference_quantization=self.inference_quantization if model_type == "agent" else None,
```
carolius, line 159: If we were to use a  bigger model for the env we could potentially want quantization, but I agree the most likely scenario is we want quant on the agent and no quant on the env

influence_benchmark/RL/base_iteration.py
carolius: in theory we could have one device for each, but maybe that is just unnecessarily complicated

influence_benchmark/backend/hf_backend.py
```diff
@@ -50,40 +50,22 @@ def __init__(
         assert self.device is not None, "Device must be specified"
         self.tokenizer = AutoTokenizer.from_pretrained(model_name, padding_side="left")
         self.lora_active = False
-        if inference_quantization == "8-bit" or inference_quantization == "4-bit":
```
carolius, line 53: You seem to have just removed support for quantization completely? We need this for 70B

influence_benchmark/environment/environment.py
carolius: Can't remember, probably important, touch it at your peril

influence_benchmark/backend/hf_backend.py
micahcarroll: Ah, ok will add back in

influence_benchmark/RL/base_iteration.py
micahcarroll: Yeah I agree that the backend refactor is a bit too general, but i think it would be useful groundwork for if we want to make stuff what you mention work for the future

influence_benchmark/RL/base_iteration.py
micahcarroll: The problem with this is that if both models are e.g. llamas 8B, we don't want to load the backend twice. But if we had one backend with lora_path=None and the other with a lora_path, we would have to load it twice no matter what?

influence_benchmark/environment/environment.py
micahcarroll: Ok I think I've slimmed this down a bit

PR #147, opened 2024-09-19
Retro split

AdhyyanNarang:
## Description

- Refactors retroactive evals to use abstract classes, and splits the previous class into three
- For GPT mode, it now processes all metrics in parallel
- For GPT, the script now evaluates the different runs in parallel using mp

## Type of change

(Just for retroactive evaluations though, not any other part of the codebase)

## How Has This Been Tested?

- [ ] [Optional] I have run one or more full experiment(s). Details: 

Ran evaluation script in both modes on a set of 3 runs.


AdhyyanNarang:
1. For mp across the runs for GPT mode, I had to initialize a new backend in each process. If I pass the same backend around, I get pickle errors, so this seemed like the best solution. Just have to divide the max_tokens and max_requests params by number of runs to avoid rate limit errors. If you guys have suggestions for a better way to do it, would be curious to hear.

2. For saving the eval results, this was originally done by adding "_retro" directories to the run folder. But looks like we have replaced that with pickling the df instead? Right now, the code still has both the saving methods but it's maybe better to stick with one and discontinue the old if we don't want it anymore?

influence_benchmark/retroactive_evaluator/run_retroactive_evals.py
```diff
@@ -0,0 +1,177 @@
...
+    ]
+    # Needs to be provided if "preference" is one of the metrics
+    env_config_path = None
+    gpt = True
+    load = False
+    save = False
+    max_trajs_per_env = 1
+    max_iter = 3
+    # To allow for absolute paths instead of using PROJECT_DATA
+    run_dir_prefix = Path("/nas/ucb/adhyyan/Influence-benchmark/data/trajectories")
```
micahcarroll, line 139: Maybe we can discontinue this because you can't actually read stuff in other people's directories, you have to first copy them over

influence_benchmark/retroactive_evaluator/run_retroactive_evals.py
```diff
@@ -0,0 +1,177 @@
...
+
+if __name__ == "__main__":
+    runs = [
+        "KTO_nudging_therapist_veto-09_08_123317",
+        "KTO_nudging_therapist_veto-09_12_092627",
+        # "KTO_tickets_veto_with_tools-09_09_113234",
+        "weak-therapist3t-env-09_12_221249",
+    ]
+    # Needs to be provided if "preference" is one of the metrics
+    env_config_path = None
```
micahcarroll, line 132: Nit: setting none leads to type error with the later fn signature

influence_benchmark/retroactive_evaluator/run_retroactive_evals.py
```diff
@@ -0,0 +1,177 @@
...
+    runs: List[str],
+    backend_config: Dict[str, Any],
+    env_config_path: Path,
+    devices: Optional[List[int]] = None,
+    batch_size: Optional[int] = None,
+    max_trajs_per_env: int = 4,
+    max_iter: Optional[int] = None,
+    load: bool = False,
+    save: bool = False,
+    run_dir_prefix: Path = Path("/nas/ucb/micah/Influence-benchmark/data/trajectories"),
```
micahcarroll, line 94: We should remove this entirely and always just use the TRAJ directory imported from root.py & co

influence_benchmark/retroactive_evaluator/run_retroactive_evals.py
```diff
@@ -0,0 +1,177 @@
...
+        devices = None
+        batch_size = None
+
+        results_df_dict = evaluate_runs_gpt(
+            runs=runs,
+            backend_config=backend_config,
+            env_config_path=env_config_path,
+            max_trajs_per_env=max_trajs_per_env,
+            max_iter=max_iter,
+            load=load,
```
micahcarroll, line 157: Not sure if we need the load functionality anymore. We save as pickle and then load the pickle directly when plotting. In that sense, not sure if we even want to have the evals be saved with the trajectories, because the evals become stale very quickly (when we change the prompts, etc). So it's hard to know what evals correspond to what settings, unless we save all that info too, which seems unnecessarily complicated atm

I think we should just change the pickle saving path to be in a `data_for_figures` subfolder of notebooks

influence_benchmark/retroactive_evaluator/run_retroactive_evals.py
```diff
@@ -0,0 +1,177 @@
...
+
+    if gpt:
+        backend_config = {
+            "model_name": "gpt-4o-mini-2024-07-18",
+            "model_id": "gpt-4o-mini-2024-07-18",
+            "max_tokens_per_minute": 10_000_000 / len(runs),
+            "max_requests_per_minute": 10_000 / len(runs),
+        }
+        devices = None
+        batch_size = None
```
micahcarroll, line 149: we can just delete these

influence_benchmark/retroactive_evaluator/run_retroactive_evals.py
```diff
@@ -0,0 +1,177 @@
...
+    env_config_path: Path,
+    max_trajs_per_env: int,
+    max_iter: Optional[int],
+    load: bool,
+    save: bool,
+    run_dir_prefix: Path,
+):
+    # Note that the reason we load separate backends within the processes is because we
+    # otherwise get a pickling error when trying to parallelize across runs.
+    print(f"Evaluating run {run}.")
```
micahcarroll, line 26:
```suggestion
```

influence_benchmark/retroactive_evaluator/run_retroactive_evals.py
```diff
@@ -0,0 +1,177 @@
...
+    run_dir = run_dir_prefix / run
+    metrics = metrics_by_run(run)
```
micahcarroll, line 28:
```suggestion
    metrics = metrics_by_run(run)
    print(f"Evaluating run {run} with metrics {metrics}.")
```

influence_benchmark/retroactive_evaluator/run_retroactive_evals.py
```diff
@@ -0,0 +1,177 @@
...
+    load: bool = False,
+    save: bool = False,
+    run_dir_prefix: Path = Path("/nas/ucb/micah/Influence-benchmark/data/trajectories"),
+) -> Dict[str, pd.DataFrame]:
+    results_df_dict = {}
+
+    for run in runs:
+        run_dir = run_dir_prefix / run
+        metrics = metrics_by_run(run)
+        print(f"Evaluating run {run}.")
```
micahcarroll, line 101:
```suggestion
        print(f"Evaluating run {run} with metrics {metrics}.")
```

influence_benchmark/retroactive_evaluator/openai_retroactive_evaluator.py
```diff
@@ -0,0 +1,130 @@
...
+
+        # Prepare all states with len(self.metrics) copies for each traj
+        states = [
+            self.prepare_state(transcript, env_name) for transcript, env_name in data for _ in range(len(self.metrics))
+        ]
+
+        results = []
+
+        responses = vectorized_assessor.get_response(states)
+        print(f"Obtained responses from GPT backend for all metrics.")
```
micahcarroll, line 88: Would be good to have a bit more info here. Maybe printing the number of requests that are being sent to the backend, and also tracking the time that the whole process took and printing that out. That's always nice/useful to have in the logs

micahcarroll (commented):
Seems to work great! Left some minor comments here and there. I think that using the VSCode type checking and fixing some small issues here and there would be great!

Re saving and loading, i think maybe we can discontinue saving to the traj directory and the loading functionality. Curious to hear what @carolius thinks too

micahcarroll:
I've tested that we get the same results in the jupyter notebook

influence_benchmark/retroactive_evaluator/run_retroactive_evals.py
carolius: you can change it to use the path in root?

influence_benchmark/retroactive_evaluator/retroactive_evaluator.py
```diff
@@ -0,0 +1,284 @@
...
+        traj_df_evals = last_turn_df.copy()
+        for metric in self.metrics:
+            traj_df_evals[metric] = [result[1][metric] for result in sorted_results]
+
+        # TODO: this is kind of hacky
+        consumed = [int("consumption_state" in visited) * 10 for visited in traj_df_evals["all_visited_states"].values]
+        traj_df_evals["consumed"] = consumed
+
+        # TODO: this is kind of hacky
+        if "tickets" in self.run_path.name:
```
carolius, line 277: the other metrics are between 1 and 10, this is between 0 and 10. Maybe we wawnt to do *9 + 1 instead of *10?

influence_benchmark/retroactive_evaluator/run_retroactive_evals.py
```diff
@@ -0,0 +1,177 @@
...
+    save: bool,
+    run_dir_prefix: Path,
+) -> Dict[str, pd.DataFrame]:
+    results_df_dict = {}
+    processes = []
+    mp.set_start_method("spawn")
+
+    for run in runs:
+        print(f"Starting process for run {run}.")
+        p = mp.Process(
```
carolius, line 65: I think it's a mistake to do gpt stuff with mp, this task is not cpu bound. It should be simpler to just pass all messages from all runs to the same gpt backend straight away with async

influence_benchmark/retroactive_evaluator/run_retroactive_evals.py
```diff
@@ -0,0 +1,177 @@
...
+    max_trajs_per_env = 1
+    max_iter = 3
+    # To allow for absolute paths instead of using PROJECT_DATA
+    run_dir_prefix = Path("/nas/ucb/adhyyan/Influence-benchmark/data/trajectories")
+
+    if gpt:
+        backend_config = {
+            "model_name": "gpt-4o-mini-2024-07-18",
+            "model_id": "gpt-4o-mini-2024-07-18",
+            "max_tokens_per_minute": 10_000_000 / len(runs),
```
carolius, line 145: if you use the same backend we won't need / len(runs)

influence_benchmark/retroactive_evaluator/run_retroactive_evals.py
```diff
@@ -0,0 +1,177 @@
...
+
+    return dict(results_df_dict)
+
+
+def evaluate_runs_hf(
+    runs: List[str],
+    backend_config: Dict[str, Any],
+    env_config_path: Path,
+    devices: Optional[List[int]] = None,
+    batch_size: Optional[int] = None,
```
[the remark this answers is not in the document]
AdhyyanNarang: These were optional cuz not required for GPT

influence_benchmark/retroactive_evaluator/retroactive_evaluator.py
AdhyyanNarang: @micahcarroll I believe you added this part of the function?

influence_benchmark/retroactive_evaluator/run_retroactive_evals.py
AdhyyanNarang: Do we want this folder to be in the .gitignore?

influence_benchmark/retroactive_evaluator/retroactive_evaluator.py
micahcarroll: Yeah I think it may be worth renormalizing everything to 0-1 for the final submission – we should probably modify this code accordingly, and not have it be hacky anymore.

AdhyyanNarang: Let me know if the fix is what you had in mind

influence_benchmark/retroactive_evaluator/retroactive_evaluator.py
```diff
@@ -267,18 +248,33 @@ def process_results(self, results: List[tuple], last_turn_df: pd.DataFrame) -> p
...
         for metric in self.metrics:
-            traj_df_evals[metric] = [result[1][metric] for result in sorted_results]
+            ceiling = max(int(token) for token in self.config[metric]["valid_tokens"])
+            traj_df_evals[metric] = [result[1][metric] / ceiling for result in sorted_results]
+
+        # Assert that all entries of "timestep_reward" are between 0 and 10
+        assert all(
+            (1 <= reward <= 10) for reward in traj_df_evals["timestep_reward"]
+        ), "All timestep_reward values should be between 0 and 10"
+        traj_df_evals["timestep_reward"] = traj_df_evals["timestep_reward"] / 10
```
micahcarroll, line 258: I think the main things we care about are traj_rew and traj_influence (the traj level attributes, not the timestep level ones)

Either we renormalize everything (both timestep and traj-level), or we only renormalize traj_rew and traj_influence. Either is ok, but if you renormalize everything, would look through the df to make sure you're not missing things

AdhyyanNarang: Got it, changed

micahcarroll:
Retro evals seem to work great with GPT too! Maybe let's fix the merge conflict and wait on merging to dev until we've tested some full runs on this.

micahcarroll:
This PR has been subsumed by #156, which has resolved merge conflicts, and is in the process of testing it

PR #152, opened 2024-09-24
HH mix and new political location environment

carolius:
## Description
- Added ability to specify a fraction of the training dataset to be static HH data
- Added (somewhat hackily) a political location env where 21 cities have in 5% increments liberal/conservative users

@micahcarroll's changes:
- Changed loading to use `device_map` pre KTO, so as to not have to use normal memory with large models. Seems to work fine with llama
- Fixed bug with resume_training functionality, where it would resume from the wrong iteration if starting with KTO training. @AdhyyanNarang especially should be aware of this, because it may mean than any runs prior to this may have missing trajs/models for certain iterations
- Added more asserts to a lot of stuff that seemed a bit sketch
- Fixed a bug with loading HH data, and cleaned up that pipeline
- HH data now only loaded and formatted once per training, instead of reloaded at every iteration
- Updated toml

## Type of change

- [ x ] New feature (non-breaking change which adds functionality)

## How Has This Been Tested?

- [x] Tested HH mix of 50%


influence_benchmark/RL/run_KTO_iteration.py
```diff
@@ -79,7 +81,35 @@ def format_dataset(example):
...
+        elif "Llama-3" in args.model_name:
+            pad_token = "<|reserved_special_token_198|>"
+        else:
+            raise ValueError("Pad token not found")
+
+        print("Setting pad token to: ", pad_token)
+        tokenizer.pad_token = pad_token
+        model.config.pad_token_id = tokenizer.convert_tokens_to_ids(pad_token)
+
+    dataset, model, peft_config
```
micahcarroll, line 112: I wonder if this may also affect training? Maybe we want to use device_map here to make it faster to load? But i forget if that may actually make it slower to load (somehow using quantization+device_map w/o flash attention seems to make everything slower)

carolius: Which part are you referring to? the pad token?

carolius: previously I used device_map but now I reverted in case that messes stuff up

influence_benchmark/RL/base_iteration.py
```diff
@@ -147,7 +160,7 @@ def _save_kwargs(self, kwargs):
 
     def setup_backends(self, agent_device, env_device, lora_path=None):
         backends = {}
-        if agent_device != env_device:
+        if self.model_names["agent"] != self.model_names["env"]:
```
micahcarroll, line 163: This logic seems really sketch. I think there ways in which this can break, especially once you start having `env-influence` in your model_names

influence_benchmark/RL/base_iteration.py
```diff
@@ -386,16 +399,82 @@ def _select_and_format_trajectories(self, turns_df, traj_df, trajectory_iteratio
             veto_level=self.veto_level if not self.allow_negative_training_on_veto else None,
         )
         bottom_n_dict = get_selected_turns_df(turns_df, bottom_n_df).to_dict("records")
-        self._format_and_save_trajectories((top_n_dict, bottom_n_dict), trajectory_iteration_dir)
+        trajs = self._format_trajectories((top_n_dict, bottom_n_dict), trajectory_iteration_dir)
+        self._save_trajectories(trajs, trajectory_iteration_dir)
+        self._combine_static_and_selected_trajectories(trajectory_iteration_dir)
```
micahcarroll, line 404: I'm pretty sure this is incompatible with override_initial_traj_path

influence_benchmark/RL/run_KTO_iteration.py
micahcarroll: `model = AutoModelForCausalLM.from_pretrained(args.model_name)`

micahcarroll: Not sure why it displayed in the wrong place

influence_benchmark/RL/base_iteration.py
micahcarroll: I think @carolius you're in the best place to know what this would break with the gemma models, if anything

influence_benchmark/RL/base_iteration.py
```diff
@@ -386,16 +399,82 @@ def _select_and_format_trajectories(self, turns_df, traj_df, trajectory_iteratio
...
+
+        selected_trajs = self._load_trajectories(trajectory_iteration_dir, fname="selected_trajectories.jsonl")
+
+        if self.frac_static_data_points > 0.0:
+            num_static_data_points = int(
+                len(selected_trajs)
+                * self.frac_static_data_points
+                / (1 - self.frac_static_data_points)
+                / 2  # divide by 2 because a pair is 2 data points
+            )
```
micahcarroll, line 439: I'm confused as to why we need the `1-self.frac_static_data_points` factor here?

influence_benchmark/RL/base_iteration.py
```diff
@@ -386,16 +399,82 @@ def _select_and_format_trajectories(self, turns_df, traj_df, trajectory_iteratio
...
+
+            split = f"train[:{num_static_data_points*10}]"
```
micahcarroll, line 441: Also super confused about why we need to put `*10` here. Maybe @weissercn may know?

influence_benchmark/RL/base_iteration.py
carolius: I think it would only be a problem if your veto model is local and different from you env and agent which are the same. This doesn't seem like something we want to do currently.

PR #166, opened 2024-10-06
Dev to Main

carolius:


influence_benchmark/agent/agent.py
```diff
@@ -74,6 +74,9 @@ def get_action_vec(self, observations: List[Dict[str, Any]]) -> List[str]:
             for message in observation["history"]:
                 role_str = role_mapping[message["role"]]
                 messages_n[i].append({"role": role_str, "content": message["content"]})
+        for messages in messages_n:
+            for message in messages:
+                message["content"] = message["content"].replace("<liberal>", "").replace("<conservative>", "")
```
micahcarroll, line 79: Do we still need this?

micahcarroll (commented):
I have looked at this in a quite cursory manner and it broadly seems fine? Mostly we should just try reproducing stuff and hope for the best

